PR #73, opened 2024-04-29
Add natural sorting for tracks and filters

guillermofbriceno:
Firstly I'd like to say I've been looking for a music player for Linux like this one for years, so thanks for your work!

At the moment, Fooyin uses ASCII-order sorting for the library, as is also the case for most other players I've used. For a music player, I believe natural sorting should be used instead. The core argument for this goes [pretty far back,](https://blog.codinghorror.com/sorting-for-humans-natural-sort-order/) but in our case, consider the current sorting of some Bach works with their commonly used Opus number:

```
BWV 1
BWV 10
BWV 100
...
```
Using `QCollator` instead, the player will use a natural sort more similar to the behavior of foobar2000:

```
BWV 1
BWV 2
BWV 3
...
```
---
I am a C programmer by trade so there might be some bad code here.

### DIFF
--- a/src/core/library/tracksort.cpp
+++ b/src/core/library/tracksort.cpp
@@ -24,6 +24,8 @@
 
 #include <ranges>
 
+#include <QCollator>
+
 namespace {
 Fooyin::ParsedScript parseScript(const QString& sort)
 {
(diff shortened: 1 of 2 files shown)


ludouzi (approved):

ludouzi:
Thanks for the contribution! I agree that natural sorting should be preferred for user-facing strings and cases such as this.

The changes look good here. I wasn't aware of QCollator before now, so thanks for bringing it to my attention. LibraryTree also uses a simple locale-aware sort and should be switched over to use QCollator, but I'll do so separately.

PR #86, opened 2024-05-15
[gui] EditableLayout: Remove margins

olib14:
This improves appearance with the KDE Breeze theme:

image

Of note, there is no longer empty space at the bottom and to the left of the artwork, which was previously present.

There is less spacing besides buttons such as those on the top toolbar, but if this was intentional to space them out, they should probably come with margins as they don't space at all against their containers (i.e. they are right against the splitter).

### DIFF
--- a/src/gui/editablelayout.cpp
+++ b/src/gui/editablelayout.cpp
@@ -197,7 +197,7 @@ struct EditableLayout::Private
         , editingContext{new WidgetContext(self, Context{"Fooyin.LayoutEditing"}, self)}
         , layoutHistory{new QUndoStack(self)}
     {
-        box->setContentsMargins(5, 5, 5, 5);
+        box->setContentsMargins(0, 0, 0, 0);
         box->addWidget(root);
 
         widgetProvider->setCommandStack(layoutHistory);


olib14:
This might regress appearance with the theme shown in the preview images, so there might be a better way to work with themes to achieve this - but I don't know how.

ludouzi:
Thanks for the suggestion! I agree it looks better with Breeze, but it might not for others. 

I've actually been planning to make the margins of each splitter configurable, including the handle size.  I think what will be best for now is to make the editable layout margin a setting (with a default of 0) so the user can change it depending on their theme and preferences.

For the margins of the controls in the toolbar, I set them all to 0 so they wouldn't be spaced too far apart when in the same splitter. I'll see about making that configurable too.

olib14:
I've been looking into theming, and it might be that themes provide a value that could be used instead of having a setting for this:

https://doc.qt.io/qt-6/qstyle.html#PixelMetric-enum

https://invent.kde.org/plasma/breeze/-/blob/master/kstyle/breezemetrics.h

I'm prepared to open another MR - what theme do you use, so I can compare?

ludouzi:
I alternate between the Otto and Nothing theme.

Actually, I've just realised the content margins of a layout are set by the style if they're not set explicitly:

> By default, QLayout uses the values provided by the style. On most platforms, the margin is 11 pixels in all directions.
https://doc.qt.io/qt-6/qlayout.html#contentsMargins

I've just tested a few styles and the default margins can vary significantly. The Breeze theme sets a default margin of 10 however, so there'd be the gap this PR was trying to solve.

It might just be a case of not setting any margins ourselves, unless the user overrides it in the setting.

olib14:
Where does the spacing between the widgets come from? Since there isn't any in Breeze, it's probably a clue for where to look. I'll investigate this myself later.

ludouzi:
I'm pretty sure the spacing between widgets comes from the QSplitter handle size, which is also determined based on the current style. Yeah, this needs more research.

ludouzi:
Yeah, spacing is provided by the `QStyle::PM_SplitterWidth` pixel metric. For Breeze this is just 1, but for Otto it's 10.

olib14:
It's better, but still not ideal. Changes to Breeze in Plasma 6 for the frameless appearance are supposed to work out of the box, but I think the use of a layout + splitters (instead of say, QDockWidgets in a QMainWindow) to organise content is problematic here.

I'll keep thinking about what can be done better. I'll also bug some of the KDE folks and see if they have any thoughts.

PR #97, opened 2024-05-18
Add option to disable ALSA with PipeWire as a default

eclipseo:


CMakeLists.txt
```diff
@@ -128,7 +129,9 @@ set_package_properties(Qt6Network PROPERTIES TYPE REQUIRED)
 set_package_properties(Qt6Svg PROPERTIES TYPE REQUIRED)
 
 find_package(Taglib REQUIRED taglib>=1.12)
+if(BUILD_ALSA)
 find_package(ALSA REQUIRED)
+endif()
```
ludouzi, line 134: You'll need to add the definition you're using here so it can be found:

`add_definitions(-DBUILD_ALSA_ENABLED)`

ludouzi (changes requested):

CMakeLists.txt
eclipseo: Ok fixed.

ludouzi (approved):
Changes look fine. I'd like to avoid preprocessor directives if possible, but in this case it would require changing how a default audio output is decided upon and loaded, so I'll save that for another time.

PR #106, opened 2024-05-20
[gui] Forward scrolling on volume button to slider

olib14:
This means that the icon can be scrolled on directly, rather than having to move to the slider.

Forwarding the event means that scrolling behaves identically as on the slider directly.

### DIFF
--- a/src/gui/controls/volumecontrol.h
+++ b/src/gui/controls/volumecontrol.h
@@ -21,6 +21,8 @@
 
 #include <gui/fywidget.h>
 
+#include <QWheelEvent>
+
 namespace Fooyin {
 class ActionManager;
 class SettingsManager;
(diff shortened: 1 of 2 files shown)


src/gui/controls/volumecontrol.h
```diff
@@ -36,6 +38,8 @@ class VolumeControl : public FyWidget
     QString name() const override;
     QString layoutName() const override;
 
+    void wheelEvent(QWheelEvent *event) override;
```
ludouzi, line 41:
```suggestion
    void wheelEvent(QWheelEvent* event) override;
```

ludouzi (changes requested):
Thanks. This is a definite improvement. 

Just a small formatting adjustment needed, following the projects .clang-format.

PR #118, opened 2024-05-26
widgets: add lyrics widget

g-fb:


ludouzi (approved):
Looks good here. 

I've already got an auto-scrolling lyrics widget planned with support for synchronised lyrics, but this is a good starting point for now.

PR #168, opened 2024-07-02
Added Dutch translation

Vistaus:
🙂


ludouzi:
All looks good here. It'd be best to use [Weblate](https://hosted.weblate.org/projects/fooyin/) for any updates in the future, but great work on what looks to be a full translation.

Thanks!

Issue #89, opened 2024-05-15
Plugin request: VGM plugins


keenanweaver:
Shed a tear and quietly uttered "it's beautiful" when I discovered fooyin this morning. Thank you for making this!

My primary use-case for foobar2000 when I used Windows was video game music playback. It has a lot of plugins and is pretty much the de facto way to listen to VGM on Windows. It'd be wonderful if fooyin could replicate that experience on Linux. [Here are the plugins that foobar has](https://www.foobar2000.org/components/tag/game+music).

- [ ] [2SF Decoder](https://www.foobar2000.org/components/view/foo_input_vio2sf)
- [ ] [AdPlug](https://www.foobar2000.org/components/view/foo_input_adplug)
- [ ] [ASAP Decoder](https://www.foobar2000.org/components/view/foo_asap)
- [ ] [foo_input_vgm](https://ux.getuploader.com/foobar2000/) More info: https://vgmrips.net/forum/viewtopic.php?p=5638#p5638
- [ ] [Game Music Emu Decoder](https://www.foobar2000.org/components/view/foo_input_gme)
- [ ] [GSF Decoder](https://www.foobar2000.org/components/view/foo_input_gsf)
- [ ] [Hively Tracker Decoder](https://www.foobar2000.org/components/view/foo_input_hvl)
- [ ] [Lunar 2 PCM Decoder](https://www.foobar2000.org/components/view/foo_lunar2)
- [ ] [MIDI Player](https://www.foobar2000.org/components/view/foo_midi)
- [ ] [NCSF Decoder](https://www.foobar2000.org/components/view/foo_input_ncsf)
- [ ] [OpenMPT Module Decoder](https://www.foobar2000.org/components/view/foo_openmpt54)
- [ ] [PMD Decoder](https://www.foobar2000.org/components/view/foo_input_pmd)
- [ ] [PSF Decoder](https://www.foobar2000.org/components/view/foo_psf)
- [ ] [QSF Decoder](https://www.foobar2000.org/components/view/foo_input_qsf)
- [ ] [SID Decoder](https://www.foobar2000.org/components/view/foo_sid+%28x64%29)
- [ ] [SNESAPU Input](https://www.foobar2000.org/components/view/foo_snesapu)
- [ ] [SSF/DSF Decoder](https://www.foobar2000.org/components/view/foo_input_ht)
- [ ] [USF Decoder](https://www.foobar2000.org/components/view/foo_input_usf)
[...]

PR #193, opened 2024-07-22
[plugins] Add a libvgm input/decoder

ludouzi:
Adds support for decoding the following file types:

- vgm
- vgz
- s98
- dro
- gym

Track loop count is configurable, and track numbers will be guessed based on the filename (can be changed). Other properties like fade duration should be made configurable, but we'll have to decide on the best way to deal with the duration changing before then.

Addresses #89


diizzyy:
Any plans on supporting this as an external dependency instead of bundling it?

ludouzi:
It could be handled similar to KDSingleApplication where we use the system library if found, and use the bundled library if not. I'll look into it.

diizzyy:
Sounds great, thanks!

PR #220, opened 2024-08-11
Fix fade steps calculation

lstrsrt:
Previously, fade-out and fade-in durations were doubled.


ludouzi:
Thanks. Looks like that was accidentally included with f8e2e994edee6f7278eaaf3808b81167a3aa45ad while I was testing changes to fading.

PR #221, opened 2024-08-11
Always reinit decoder when track was stopped

lstrsrt:
Without this patch, after reaching the end of a playlist, hitting Play again results in occasional segfaults/crashes. I wasn't 100% able to figure out why it sometimes doesn't happen, but this seems to fix it.

### DIFF
--- a/src/core/engine/audioplaybackengine.cpp
+++ b/src/core/engine/audioplaybackengine.cpp
@@ -199,22 +199,24 @@ void AudioPlaybackEngine::play()
         return;
     }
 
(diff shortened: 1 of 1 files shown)


ludouzi (approved):

ludouzi:
Thanks. Looks like crashes occur because TrackStatus::End is set if we get to the end of the playlist without manually stopping, but this state isn't used when checking if we need to reinit the decoder. 

The fact that it occasionally doesn't happen means we stop playback in a buffered state which I'll look into.

PR #222, opened 2024-08-12
Clamp fade durations and pause next track if needed

lstrsrt:
When playing/pausing/stopping near the end of a track with fading enabled, the fade duration can exceed the
remaining track duration. The current behavior is to cut off the fade before it has fully finished and then start the next track, even if the user pressed pause and expects playback to stop.

This patch clamps fade durations to the remaining track duration and, if pause or stop was pressed, pauses
when reaching the next song.

### DIFF
--- a/include/core/engine/enginecontroller.h
+++ b/include/core/engine/enginecontroller.h
@@ -40,6 +40,8 @@ public:
         : QObject{parent}
     { }
 
+    [[nodiscard]] virtual PlaybackState engineState() const = 0;
+
     /** Returns a list of all output names. */
     [[nodiscard]] virtual OutputNames getAllOutputs() const = 0;
 
--- a/src/core/application.cpp
+++ b/src/core/application.cpp
@@ -184,7 +184,8 @@ void ApplicationPrivate::setupConnections()
     m_settings->subscribe<Settings::Core::Shutdown>(m_self, [this]() {
         savePlaybackState();
 
-        if(m_playerController->playState() == Player::PlayState::Playing) {
+        const auto state = m_engine.engineState();
+        if(state == PlaybackState::Playing || state == PlaybackState::Fading) {
             QObject::connect(&m_engine, &EngineController::finished, m_self, Application::quit);
             m_playerController->stop();
         }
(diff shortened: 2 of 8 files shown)


ludouzi:
Nice work. 

From a quick glance, a potential issue is that some VGM decoders support infinite looping, meaning the track will continue playing past it's duration. For those, I'm guessing the first call to play will keep it paused. This may also apply to tracks which don't have a valid duration, wherein the maximum value of uint64_t is used.

To cover those cases, perhaps an additional bool can be used in changeTrack so we know whether we've actually reached the end and need to pause the next.

lstrsrt:
Yeah, actually this is a lot easier than I thought -- if we were fading in changeTrack, just set m_pauseNextTrack to true. m_pauseNextTrack is slightly misnamed because we may also want to pause the current track instead (if the track is being repeated, I think it should be paused at the start after the fade-out -- that's the behavior of this patch at least.)

About the clamping behavior in these two cases:
For an invalid m_duration, we don't know when the track will end, so we should probably keep the duration as-is. 
For the VGM case, if m_lastPosition can be > m_duration, `remaining = m_duration - (m_lastPosition % m_duration)` is probably enough (?).

ludouzi:
I've just tried this out - pausing works fine, but stopping at the end will lead to a paused state on pressing play.

Another issue somewhat related is seeking while fading at the end; it will currently always change to the next track once the fade ends.

This might be fixed/further simplified by just preventing the change to the next track altogether by adding a check in _onRendererFinished_. Then changes wouldn't need to be made in _changeTrack_ or _play_. I'll do some testing of this.

ludouzi:
I've implemented the above change and it fixes several issues. When pausing/stopping at the end, the track will fade and then remain paused/stopped at the end. On starting playback, the next track will play if paused, or the current track will be restarted if it was stopped.

Issues fixed:

- Seeking during fading at the end
- Stopping during a faded pause
- Fade restarting on app close

Ideally we would end paused/stopped at the start of the next track + any remaining fade duration, but there are problems with this like the above. Once work starts on DSP plugins, we'll need to be able to decode multiple tracks simultaneously to handle crossfading, so this can be revisited then.

lstrsrt:
Sounds good; are these changes going to be merged from base or should I keep this PR open?

PR #232, opened 2024-08-18
Allow changing album cover alignment

lstrsrt:
This adds an action group to the album cover context menu allowing the user to change the album cover alignment. The alignment is also saved & loaded from the layout configuration.


ludouzi (approved):
Thanks, all looks good here.

PR #237, opened 2024-08-19
Use S-curve for longer fade durations

lstrsrt:
This implements a simple S-curve fading algorithm for fade durations over 1000ms. The linear algo makes the end of a fade-out/start of a fade-in sound very abrupt, which this patch fixes.

The formula is not perfect, there is probably a more efficient/better version but this sounds pretty good to me as a start.
I've also changed the code so that pausing/unpausing repeatedly does not reset the fade steps every time.


ludouzi (approved):

ludouzi:
Thanks. This definitely sounds better to me as well.

PR #240, opened 2024-08-22
Switch playlist when scrolling over playlist tabs

lstrsrt:
Previously, scrolling over playlist tabs already visually changed the active tab, but did not update the playlist view. This fixes it, and adds wrap-around when scrolling at the first or last playlist.

### DIFF
--- a/include/utils/widgets/editabletabbar.h
+++ b/include/utils/widgets/editabletabbar.h
@@ -51,6 +51,7 @@ signals:
     void tabTextChanged(int index, const QString& text);
 
 protected:
+    bool event(QEvent* event) override;
     void mouseDoubleClickEvent(QMouseEvent* event) override;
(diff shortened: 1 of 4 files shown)


ludouzi:
Also, by moving the wrap-around implementation to EditableTabBar, we could just emit tabBarClicked on scroll, which already handles changing the playlist in the playlist view. I'll make some changes in a bit.

ludouzi:
> I'm guessing because it's propagated differently and doesn't have to talk to the playlist controller.

Yeah, the Tab Stack uses a QTabWidget which contains a QTabBar and and a QStackedWidget, so it handles switching internally.

> I found that switching to bigger playlists (> 500 items) has a noticeable delay when compared to smaller playlists.

The playlist model is built asynchronously, so a number of tracks are used to preload it (currently 2000). This number could be adjusted to improve that, or model updates could be made in batches.

PR #243, opened 2024-08-26
Scroll ExpandedTreeView if selected item is not fully visible

lstrsrt:
Previously, selecting an item which was only barely visible did not scroll the playlist or the queue viewer, which felt a bit irritating to me. I also simplified the logic in cmpItemsPlaylistItems; I can do a separate PR for this if needed but it's a very small change.

### DIFF
--- a/src/gui/playlist/playlistmodel.cpp
+++ b/src/gui/playlist/playlistmodel.cpp
@@ -75,10 +75,6 @@ bool cmpItemsPlaylistItems(Fooyin::PlaylistItem* pItem1, Fooyin::PlaylistItem* p
         }
     }
 
-    if(item1->row() == item2->row()) {
-        return false;
-    }
-
     return reverse ? item1->row() > item2->row() : item1->row() < item2->row();
 };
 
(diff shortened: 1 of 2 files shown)


ludouzi:
Thanks for the simplification in cmpItemsPlaylistItems. That was originally used for sorting so strict weak ordering was necessary.

For scrolling to the selected item in ExpandedTreeView, I believe the same thing can be accomplished by changing a single line.

When the current index changes in a QAbstractItemView, scrollTo is automatically called with EnsureVisible. It will consider the index visible if any part of the visualRect intersects the viewport:

https://github.com/fooyin/fooyin/blob/f539d19ce466bd670a0221c4aaf2d51f706787c0/src/utils/widgets/expandedtreeview.cpp#L3208

Changing the QRect::intersects call to QRect::contains will require the viewport to fully contain the visualRect instead.

ludouzi:
Yeah, that's because _scrollTo_ is called directly by _currentChanged_ on key press. For scrolling after a selection, it's posted to the event loop. Though that delay matches the behaviour of the default Qt item views so it shouldn't be an issue.

PR #258, opened 2024-09-02
Fix playback not pausing while muted

lstrsrt:
We had a division by zero in AudioRenderer::pause if fadeLength was 0, which turned m_volumeChange into -inf (if volume was not muted) and nan (if volume was muted). In the second condition, we would never call pauseOutput.

This is fixed by making the non-fade pause/play functions return early, since they don't care about the fading variables or the fade timer anyway.

We can additionally skip fading in AudioPlaybackEngine::pause if the volume is muted, like it's already done in ::stop. If nothing is being played, there's no use in fading anything out either.

### DIFF
--- a/src/core/engine/audioplaybackengine.cpp
+++ b/src/core/engine/audioplaybackengine.cpp
@@ -253,7 +253,7 @@ void AudioPlaybackEngine::pause()
         return;
     }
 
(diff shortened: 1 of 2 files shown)


ludouzi (approved):
Thanks, nice catch.

PR #259, opened 2024-09-03
[utils] Timer: Add missing include for sstream

diizzyy:
Fixes build issue on FreeBSD 14.0

error: implicit instantiation of undefined template 'std::basic_ostringstream<char>


ludouzi (approved):
Thanks 👍

PR #262, opened 2024-09-05
Add ReplayGain playback support and settings page

lstrsrt:
This patch adds support into the playback engine for playing tracks with ReplayGain tags.
It also adds a settings page with an enablement option (disabled by default), a track/album gain radio selection and a pre-amp slider.

Some notes:
* There is a delay of a few seconds before anything happens when any RG setting is changed. Some kind of Linux audio driver issue? I've had this happen on Pipewire and ALSA.
* The implementation guide recommends dithering after scaling the audio, which I've not done here (yet)
* Currently we always use the track peak for clipping prevention, and it's done by simply lowering the gain if needed. I feel like compression just sounds too ugly to be used, but that could be made a configurable option.
* fb2k also supports RG "in playback order", I don't know what this means so I didn't add it.

I'm sure all these problems can be fixed in time.

### DIFF
--- a/include/core/coresettings.h
+++ b/include/core/coresettings.h
@@ -57,7 +57,9 @@ enum CoreSettings : uint32_t
     ExternalSortScript      = 18 | Type::String,
     Shutdown                = 19 | Type::Bool,
     StopAfterCurrent        = 20 | Type::Bool,
-
+    ReplayGainEnabled       = 21 | Type::Bool,
+    ReplayGainType          = 22 | Type::Int,
+    ReplayGainPreAmp        = 23 | Type::Int,
 };
 Q_ENUM_NS(CoreSettings)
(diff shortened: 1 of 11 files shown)


ludouzi:
Thanks for all your work on this.

>There is a delay of a few seconds before anything happens when any RG setting is changed

That's due to the buffer queue in AudioRenderer. It's directly correlated to the value of `Settings->Playback->Output->Buffer length`. To fix that, the buffers in the queue will also need to be scaled.

> fb2k also supports RG "in playback order", I don't know what this means so I didn't add it.

I believe that setting switches between track-gain and album-gain depending on the playback order i.e. shuffle (tracks) will use track-gain, default will use album-gain.

ludouzi (approved):

PR #264, opened 2024-09-08
Apply ReplayGain in AudioRenderer

lstrsrt:
Propagate gainScale to the audio renderer. This makes the delay when changing RG settings a lot shorter.

### DIFF
--- a/src/core/engine/audiorenderer.cpp
+++ b/src/core/engine/audiorenderer.cpp
@@ -19,8 +19,12 @@
 
 #include "audiorenderer.h"
 
+#include "internalcoresettings.h"
+
+#include <core/coresettings.h>
 #include <core/engine/audiobuffer.h>
 #include <core/engine/audiooutput.h>
+#include <utils/settings/settingsmanager.h>
 #include <utils/threadqueue.h>
 
 #include <QBasicTimer>
@@ -35,8 +39,9 @@ using namespace std::chrono_literals;
 constexpr auto FadeInterval = 10;
 
 namespace Fooyin {
-AudioRenderer::AudioRenderer(QObject* parent)
+AudioRenderer::AudioRenderer(SettingsManager* settings, QObject* parent)
     : QObject{parent}
+    , m_settings{settings}
     , m_volume{0.0}
     , m_gainScale{1.0}
     , m_bufferSize{0}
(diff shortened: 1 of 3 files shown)


ludouzi:
I'm thinking it may be better to apply only in the renderer, as otherwise the gain will be applied twice. It should also be applied to the temporary buffer just before it's sent to the output driver, as the buffers in the renderer queue may not be fully consumed each call.

I'm also experiencing artifacts when scaling non-float samples, so AudioBuffer::scale will also need looking at.

I'll make some changes in a bit.

ludouzi:
Looks like the artifacts I'm experiencing are due to overflows when scaling signed samples below 32bit. Casting to a higher precision (int32_t) should fix that.

src/core/engine/audiorenderer.cpp
```diff
@@ -379,6 +386,38 @@ void AudioRenderer::updateInterval()
...
+            peak        = m_currentTrack.replayGainTrackPeak();
+            break;
+        case(ReplayGainType::Album):
+            m_gainScale = std::pow(10.0, m_currentTrack.replayGainAlbumGain() / 20.0);
+            peak        = m_currentTrack.replayGainAlbumPeak();
+            break;
+    }
+
+    const auto preamp = static_cast<float>(m_settings->value<Settings::Core::ReplayGainPreAmp>());
+    m_gainScale *= std::pow(10.0, preamp / 20.0);
```
lstrsrt, line 411: We can add preamp earlier (in the first pow) and reduce to

`m_gainScale = std::pow(10.0, (m_currentTrack.replayGain{Album,Track}Gain() + preamp) / 20.0)`

to save the CPU some math work

ludouzi: Good call. It shouldn't matter too much as the method is only called on track change/changing settings, but it does simplify it a bit.

lstrsrt:
> I'm thinking it may be better to apply only in the renderer, as otherwise the gain will be applied twice.
I may have misworded the commit message, but this was already done in the original commit (I removed the buffer.scale() call.) Moving the rest of the calculation for the renderer is cleaner overall though.

For the overflow issue, does that mean calling scale<int32_t>()?

ludouzi:
> I may have misworded the commit message, but this was already done in the original commit

Yeah, I just noticed that. The renderer runs in a separate thread so it's safer to apply there as well.

> For the overflow issue, does that mean calling scale<int32_t>()?

It will need to be read into it's actual type first (int16_t), then cast to int32_t before scaling, and then cast back. It would probably be a good idea to clamp it to the min/max value too.

lstrsrt:
Clamping to max would mean clipping, right? Wouldn't it be better to figure out the loudest sample and clamp the scaling factor to (max / peak) so there's no overflow in the first place? Actually, I thought track peak existed for that reason.

ludouzi:
Yeah, track peak will prevent clipping for tracks with replaygain info, but not for those without. It'd probably be a good idea in this case to follow other players and split the preamp setting into two settings: 1 for when replaygain info exists and another for when it doesn't.

ludouzi:
I've done a bit more research into this and it turns out that all decoders in fb2k 2.0 output float64/double audio. That would resolve any potential overflow issues and would also make it easier to implement DSP plugins down the line as they wouldn't have to account for different sample formats.

I'll look into that approach.

lstrsrt:
I see that you already implemented this, but would float32 also have been enough? Of course precision is higher with float64, but memory usage doubles in comparison..

ludouzi:
True, but I didn't notice any increase in memory usage during testing, and as no device supports float64, it's down-converted soon after any gain is applied.

One of the reasons fb2k moved to using float64 is that is supports bitexact passthrough of all formats, and the fact that float32 can't fit all values of int32_t.

PR #267, opened 2024-09-10
Add volume slider dB tooltip

lstrsrt:
This patch adds an optional dB tooltip to the volume slider which can be enabled from the context menu.

https://github.com/user-attachments/assets/30d5837d-42d7-4a9e-bf64-4a8db9a33da5

The positioning code is not perfect; I had some problems with widget coordinates/positioning. For horizontal sliders, I used modified positioning code from seekbar.cpp. It works well enough, but the tooltip position is clamped too late on the sides and it jumps to the left when mute is reached (especially noticeable with longer sliders.) For vertical sliders, I improvised something to start with, but the tooltip is too far away from the slider on the x axis (and again there are weird jumps to the left.)

A side issue I noticed while making this is that in icon-only mode, the vertical slider always appears on the left-most position of the widget space instead of over the icon:

image

### DIFF
--- a/src/gui/controls/volumecontrol.cpp
+++ b/src/gui/controls/volumecontrol.cpp
@@ -181,7 +181,7 @@ void VolumeControlPrivate::showVolumeMenu() const
     // Only display volume slider above icon if it won't clip above the main window.
     const bool displayAbove = (yPosToWindow - menuHeight) > 0;
 
-    static constexpr int x = -15;
+    const int x            = (m_self->width() / 2) - (m_volumeMenu->sizeHint().width() / 2);
     const int y            = displayAbove ? (-menuHeight - 10) : (m_self->height() + 10);
 
     const QPoint pos(m_self->mapToGlobal(QPoint{x, y}));
--- a/src/gui/controls/volumecontrol.cpp
+++ b/src/gui/controls/volumecontrol.cpp
@@ -145,8 +145,9 @@ void VolumeControlPrivate::changeDisplay(VolumeControl::Options options, bool in
     }
 
     if(options & VolumeControl::Tooltip) {
-        QObject::connect(m_volumeSlider, &LogSlider::sliderMoved, [this](int value) { updateToolTip(value); });
-        QObject::connect(m_volumeSlider, &LogSlider::sliderReleased, [this]() {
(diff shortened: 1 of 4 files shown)


src/gui/controls/volumecontrol.cpp
```diff
@@ -138,6 +143,15 @@ void VolumeControlPrivate::changeDisplay(VolumeControl::Options options, bool in
...
     }
+
+    if(options & VolumeControl::Tooltip) {
+        QObject::connect(m_volumeSlider, &LogSlider::sliderMoved, [this](int value) { updateToolTip(value); });
+        QObject::connect(m_volumeSlider, &LogSlider::sliderReleased, [this]() {
+            if(m_toolTip) {
+                m_toolTip->deleteLater();
+            }
+        });
+    }
```
ludouzi, line 154: You'll want to pass a context object as the 3rd argument for the connections. Using the same sender `m_volumeSlider` will do.

src/gui/controls/volumecontrol.cpp
```diff
@@ -206,6 +220,54 @@ void VolumeControlPrivate::updateDisplay(double volume) const
...
+        m_toolTip->show();
+    }
+
+    constexpr auto MinVolumeLog10 = -2;
+    if(value > (MinVolumeLog10 * m_volumeSlider->scale())) {
+        m_toolTip->setText(QString(u8"%1 dB").arg(20 * (value / m_volumeSlider->scale()), 0, 'f', 1));
+    }
+    else {
+        m_toolTip->setText(QString(u8"-∞ dB"));
+    }
```
ludouzi, line 236: Using `QStringLiteral` for setting the tooltip text is preferred here, as it will save on allocation at runtime.

ludouzi (changes requested):

ludouzi:
Nice addition. I'll look into the clamping issue later.

For the vertical slider issue, the popup menu is positioned 15 pixels from the left of the widget:
https://github.com/fooyin/fooyin/blob/25b8e0580e3681414b4cd65b459754df8b1a4f4b/src/gui/controls/volumecontrol.cpp#L169

As the volume icon is always displayed in the centre of the widget, that can be changed to something like:
```
const int x = (m_self->width() / 2) - (m_volumeMenu->sizeHint().width() / 2);
```

ludouzi (approved):

PR #269, opened 2024-09-12
Add ffmpeg ReplayGain calculation support

lstrsrt:
This PR adds ReplayGain calculation support for tracks decoded by ffmpeg.

### DIFF
--- a/src/core/engine/ffmpeg/ffmpeginput.h
+++ b/src/core/engine/ffmpeg/ffmpeginput.h
@@ -31,8 +31,6 @@ class AudioBuffer;
 class FFmpegInputPrivate;
 class SettingsManager;
 
-Stream findAudioStream(AVFormatContext* context);
-
 class FFmpegDecoder : public AudioDecoder
 {
 public:
--- a/src/core/engine/ffmpeg/ffmpeginput.h
+++ b/src/core/engine/ffmpeg/ffmpeginput.h
@@ -21,11 +21,17 @@
 
 #include <core/engine/audioinput.h>
 
+#include "ffmpegstream.h"
+
+struct AVFormatContext;
+
 namespace Fooyin {
 class AudioFormat;
 class AudioBuffer;
 class FFmpegInputPrivate;
 class SettingsManager;
+
+Stream findAudioStream(AVFormatContext* context);
 
 class FFmpegDecoder : public AudioDecoder
 {
--- a/src/core/engine/ffmpeg/ffmpegreplaygain.cpp
+++ b/src/core/engine/ffmpeg/ffmpegreplaygain.cpp
@@ -44,70 +44,7 @@ extern "C"
 
 Q_LOGGING_CATEGORY(REPLAYGAIN, "fy.replaygain")
 
-namespace Fooyin {
-
-using Utils::printError;
-
-FFmpegReplayGain::FFmpegReplayGain(MusicLibrary* library)
-    : m_file{nullptr}
-    , m_decoder{nullptr}
-    , m_source{}
-    , m_library{library}
-{ }
-
-void FFmpegReplayGain::calculate(const TrackList& tracks, bool asAlbum)
-{
-    if(asAlbum) {
-        handleAlbum(tracks);
-    }
-    else {
-        for(auto track : tracks) {
-            if(setupTrack(track)) {
-                handleTrack(track);
-            }
-        }
-    }
-    if(m_decoder) {
-        m_decoder->stop();
-    }
-}
-
-bool FFmpegReplayGain::setupTrack(Track track)
-{
-    if(m_decoder) {
-        m_decoder->stop();
-    }
-
-    if(!track.isValid()) {
-        return false;
-    }
-
-    if(!track.isInArchive() && !QFile::exists(track.filepath())) {
-        return false;
-    }
-
-    AudioSource source;
-    source.filepath = track.filepath();
-    if(!track.isInArchive()) {
-        m_file = std::make_unique<QFile>(track.filepath());
-        if(!m_file->open(QIODevice::ReadOnly)) {
-            qCWarning(REPLAYGAIN) << "Failed to open" << track.filepath();
-            return false;
-        }
-        source.device = m_file.get();
-    }
-
-    m_decoder         = std::make_unique<FFmpegDecoder>();
-    const auto format = m_decoder->init(source, track, AudioDecoder::NoSeeking | AudioDecoder::NoInfiniteLooping);
-    if(!format) {
-        return false;
-    }
-
-    m_format = format.value();
-
-    return true;
-}
-
+namespace {
 struct FilterContextDeleter
 {
     void operator()(AVFilterContext* filter) const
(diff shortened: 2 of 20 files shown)


lstrsrt:
As for calculating values outside of ffmpeg, I can imagine something like a `ReplayGainCalculator` interface, which `FFmpegReplayGain` would derive from... and tracks using other decoders would have a different implementation that uses some generic library.

lstrsrt:
The main feature is sort of complete now (with caveats below) so I'm opening for review @ludouzi 

Unfortunately, calculating values for an album only works when all files have the same format. There is also no progress reporting and of course it only works for files supported by ffmpeg.

ludouzi:
Thanks for all the work on this so far. I'll go through it later.

As v0.7.0 will be released in the next few days, I think it'd be best to wait to merge until the next release so there's more time for testing. I'm also working on a more generic solution using libebur128 which I'd like to finish so calculating rg info covers all formats.

diizzyy:
Why not reuse ffmpeg instead of introducing another library? If you decide to use another library despite that I would highly recommend that https://github.com/sdroege/ebur128 is used as it's much faster (it uses the same API as libebur128).

ludouzi:
Because then rg calculation will only be possible for the formats supported by ffmpeg. I'm thinking of moving all rg calculation to a plugin as well so it will be optional.

I'll look into ebur128. The ffmpeg filter implementation uses a modified libebur128 anyway, so I'll do some profiling when I get to that.

diizzyy:
I'm going to assume that all input plugins output raw audio so just feed the output of unsupported formats?

ludouzi:
We could try feeding the audio data into AVFrame's in that case. I'll see if that's feasible.

src/core/engine/ffmpeg/ffmpegreplaygain.cpp
```diff
@@ -0,0 +1,383 @@
...
+        }
+    }
+
+    if(!finishFilter(m_albumFilter.filterContext)) {
+        return;
+    }
+
+    const auto albumResult = extractRGValues(m_albumFilter.filterGraph.get());
+
+    for(size_t i = 0; auto track : album) {
```
ludouzi, line 374: Track should be accessed by reference after copying the input TrackList.

src/core/engine/ffmpeg/ffmpegreplaygain.cpp
```diff
@@ -0,0 +1,383 @@
...
+        track.setRGTrackGain(static_cast<float>(trackResults[i].gain));
+        track.setRGTrackPeak(static_cast<float>(trackResults[i].peak));
+        track.setRGAlbumPeak(static_cast<float>(albumResult.peak));
+        track.setRGAlbumGain(static_cast<float>(albumResult.gain));
+        m_library->writeTrackMetadata({track});
```
ludouzi, line 379: `writeTrackMetadata` is quite expensive, so it should be moved outside of the loop and a single call used for the entire album.

src/core/engine/ffmpeg/ffmpegreplaygain.cpp
```diff
@@ -0,0 +1,383 @@
...
+        return;
+    }
+
+    if(!m_albumFilter.filterContext || !m_albumFilter.filterGraph) {
+        return;
+    }
+
+    std::vector<ReplayGainResult> trackResults;
+
+    for(auto track : album) {
```
ludouzi, line 362: This should be iterated by const reference.

src/core/engine/ffmpeg/ffmpeginput.h
```diff
@@ -21,11 +21,17 @@
...
+#include "ffmpegstream.h"
+
+struct AVFormatContext;
+
 namespace Fooyin {
 class AudioFormat;
 class AudioBuffer;
 class FFmpegInputPrivate;
 
+Stream findAudioStream(AVFormatContext* context);
```
ludouzi, line 33: I think this would be better suited in ffmpegutils

ludouzi (changes requested):

src/core/engine/ffmpeg/ffmpegreplaygain.cpp
```diff
@@ -0,0 +1,383 @@
...
+    if(asAlbum) {
+        p->handleAlbum(tracks);
+    }
+    else {
+        for(auto track : tracks) {
+            if(p->setupTrack(track, p->m_trackFilter)) {
+                const auto result = p->handleTrack(false);
+                track.setRGTrackGain(static_cast<float>(result.gain));
+                track.setRGTrackPeak(static_cast<float>(result.peak));
+                p->m_library->writeTrackMetadata({track});
```
ludouzi, line 162: Same as the for the comment in handleAlbum, this should be moved outside of the loop and a single call made.

lstrsrt:
Using peak=sample for the filter speeds up gain calculation by at least 2x... peak=true seems more accurate though. Could be turned into a setting?

ludouzi:
Calculating true peak would definitely be useful as an optional setting, yeah.

ludouzi (approved):

ludouzi:
After a lot of benchmarking over the past few days, I think it'd be best to mainly rely on libebur128/ebur128 for this. 

The FFmpeg filter comes with a lot of limitations - including a lack of support for format changes. It's also very slow, particularly when calculating true peaks.

Using parallel processing improves this significantly when scanning per-track, but I don't think it's possible to use the same approach for album scanning without huge memory usage (>1GB). It's not designed for processing a large number of files simultaneously.

For the set of 50 tracks I used for testing, it takes an average of 02:40 to scan as a single album. Using `libebur128` (with parallel processing) brings this down to 00:20. `ebur128` may reduce this even further.

lstrsrt:
8x really is a huge difference... it makes me wonder what the bottleneck in ffmpeg is. But in that case I would also go for (lib)ebur128.

ludouzi:
It's mainly due to (the lack of) parallel processing. I think the album filter graph expects sequential frames, but decoding multiple tracks simultaneously means adding frames from different sources randomly which I wasn't able to get working without crashing. 

Doing the same with libebur128 is trivial as it just expects the raw data and calculating the album gain can be done at the end by passing all the track states.

diizzyy:
Mainly as data point but the numbers seems to be a bit different to what I'm seeing on my box,

FreeBSD 14.1
LLVM/Clang: 18.1.5
Rust: 1.81.0

FFmpeg 6.1.2, with --enable-lto --enable-optimizations
CFLAGS= -march=znver4

rsgain 3.5.2 (uses FFmpeg for decoding)
CFLAGS= -O2 -march=znver4

ebur128 0.1.9 (rust)
CARGO_PROFILE_RELEASE_LTO="true"
CARGO_PROFILE_RELEASE_PANIC="abort"
CARGO_PROFILE_RELEASE_CODEGEN_UNITS=1
-C target-cpu=znver4

libebur128 1.2.6
CFLAGS= -O2 -march=znver4

=== True peak ===

FFmpeg 
```
/usr/bin/time -l -h ffmpeg -v quiet -stats -i sample.m4a -filter_complex ebur128=peak=true -f null -
        37.55s real             36.86s user             2.36s sys
     66036  maximum resident set size
```

rsgain + ebur128 (Rust)

```
/usr/bin/time -l -h rsgain custom -s s -t sample.m4a
        15.42s real             15.39s user             0.02s sys
     64072  maximum resident set size
```

rsgain + libebur128
```
/usr/bin/time -l -h rsgain custom -s s -t sample.m4a
        35.69s real             35.68s user             0.01s sys
     65324  maximum resident set size
```

=== Sample peak ===

FFmpeg (sample peak mode)
```
/usr/bin/time -l -h ffmpeg -v quiet -stats -i sample.m4a -filter_complex ebur128=peak=sample -f null -
        16.01s real             15.69s user             1.72s sys
     65380  maximum resident set size
```

rsgain + ebur128 (Rust)
```
/usr/bin/time -l -h rsgain custom -s s sample.m4a
        8.18s real              8.15s user              0.02s sys
     64080  maximum resident set size
```

rsgain + libebur128
```
/usr/bin/time -l -h rsgain custom -s s sample.m4a
        14.20s real             14.17s user             0.03s sys
     65320  maximum resident set size
```

=== No peak mode ===
Not really sure what that means but that's what FFmpeg's documentation calls it

FFmpeg
```
/usr/bin/time -l -h ffmpeg -v quiet -stats -i sample.m4a -filter_complex ebur128 -f null -
        14.02s real             13.72s user             1.84s sys
     65388  maximum resident set size
```

Sample file:
```
mediainfo sample.m4a
General
Complete name                            : sample.m4a
Format                                   : MPEG-4
Format profile                           : Apple audio with iTunes info
Codec ID                                 : M4A  (M4A /isom/iso2)
File size                                : 213 MiB
Duration                                 : 3 h 3 min
Overall bit rate mode                    : Variable
Overall bit rate                         : 162 kb/s
Writing application                      : Lavf60.16.100

Audio
ID                                       : 1
Format                                   : AAC LC
Format/Info                              : Advanced Audio Codec Low Complexity
Codec ID                                 : mp4a-40-2
Duration                                 : 3 h 3 min
Bit rate mode                            : Variable
Bit rate                                 : 160 kb/s
Maximum bit rate                         : 163 kb/s
Channel(s)                               : 2 channels
Channel layout                           : L R
Sampling rate                            : 44.1 kHz
Frame rate                               : 43.066 FPS (1024 SPF)
Compression mode                         : Lossy
Stream size                              : 210 MiB (99%)
Default                                  : Yes
Alternate group                          : 1
```

ludouzi:
Yeah, ffmpeg is perfectly fine for scanning a single/few files. It just can't compete with libebur128/ebur128 in fooyin's case when you want to calculate the album data for 100+ tracks without the ability to process those files in parallel.

diizzyy:
What I found a bit to be different and interesting was both performance and memory usage numbers,

Looking at rsgain it launches multiple workers/jobs per directory so I guess you essentially could call it scanning directories in parallel, https://github.com/complexlogic/rsgain?tab=readme-ov-file#multithreaded-scanning

Running 5 workers resulted in,
    237140  maximum resident set size

At least resident memory usage is appears to be somewhat expected, compared to a single worker. Given your comments above it seems to be a different approach than creating a separate instance for each file to scan?

ludouzi:
Ah, I see. Yeah, I'm talking about a separate worker for each file, rather than one for each album/directory. 

When testing multiple workers in the ffmpeg scanner, I had to store all frames of all the files in the album and then process their album data, hence the high memory usage. 

Implementing multiple workers for each album/directory like rsgain should be possible, as they would share the same album filter. And it would avoid the high memory usage. It still wouldn't be as fast as a separate instance for each file, but I'll see about adding that.

PR #290, opened 2024-09-24
Fix formatting durations equal to or longer than 1 day

lstrsrt:
This fixes the duration display in Selection Info when selecting a lot of songs in a long playlist (previously it would get reset to 0 after 23:59:59.)

The "d" suffix is not very common but it's what foobar does and avoids translation.


ludouzi:
Ha, thanks for noticing that. 

`msToStringExtended` was originally used for `%duration%`, though I don't know why it was used for the selection info as well.

I don't see any reason to not use `msToString` there, and delete `msToStringExtended` as it's not used anywhere else.

lstrsrt:
Okay, improved version is out.

ludouzi (approved):

PR #291, opened 2024-09-24
Add query syntax to ScriptParser

ludouzi:
Adds the ability to filter a TrackList using a query-based syntax on top of the normal scripting syntax. All expressions are performed case-insensitively.

### Operators

Operators are typically used to compare a metadata field to something else. 
A few of them have an equivalent keyword which can be used interchangeably.

| Operator | Syntax          | Description                                      | Keyword Equivalent |
|----------|-----------------|--------------------------------------------------|--------------------|
| `:`      | field : string  | **field** contains **string**                    | HAS                |
| `=`      | field = string  | **field** is equal to **string**                 | IS                 |
| `!=`     | field != string | **field** is not equal to **string**             |                   |
| `>`      | field > number  | **field** is greater than **number**             | GREATER            |
| `<`      | field < number  | **field** is less than **number**                | LESS               |
| `>=`     | field >= number | **field** is greater than or equal to **number** |                    |
| `<=`     | field <= number | **field** is less than or equal to **number**    |                    |
| `!`      | !expression     | **expression** is not true                       | NOT                |

### Keywords

Keywords are usually used either before or after an expression, or to group expressions together.

| Keyword              | Syntax                        | Description                                                                |
|----------------------|-------------------------------|----------------------------------------------------------------------------|
| `PRESENT`            | field PRESENT                 | Returns tracks which have the metadata field **field**                     |
| `MISSING`            | field MISSING                 | Returns tracks which don't have the metadata field **field**               |
| `BEFORE`             | time1 BEFORE time2            | Returns tracks where the **time1** field is before **time2**               |
| `AFTER`              | time1 AFTER time2             | Returns tracks where the **time1** field is after **time2**                |
| `SINCE`              | time1 SINCE time2             | Returns tracks where the **time1** field is at or after **time2**          |
| `DURING`             | time1 DURING time2            | Returns tracks where the **time1** field is within the period of **time2** |
| `AND`                | expression1 AND expression2   | Returns tracks where both expressions are true                             |
| `OR`                 | expression1 OR expression2    | Returns tracks where either expression is true                             |
| `XOR`                | expression1 XOR expression2   | Returns tracks where only one of the expressions is true                   |
| `ALL`                | ALL                           | Returns all tracks                                                         |
| `SORT BY`            | SORT BY expression            | Sort tracks (ascending) by **expression**                                  |
| `SORT ASCENDING BY`  | SORT ASCENDING BY expression  | Sort tracks (ascending) by **expression**                                  |
| `SORT DESCENDING BY` | SORT DESCENDING BY expression | Sort tracks (descending) by **expression**                                 |

Expressions can be grouped using parenthesis to change operator precedence:

`(playcount>0 AND genre=rock) OR title:rock`

Groups can also be negated using the not operator `!`:

`!(playcount>1 AND genre=classical)`

Still to do:

- [x] Time-related operators
- [x] Operator or keyword for checking if metadata is missing/present
- [x] More test cases


ludouzi:
There are performance issues when using `%date%` or `%year%` in queries, and it looks like it's due to `QDateTime::fromString`.

Ideally we would also store those fields as msecs since epoch. Looking into using std::chrono to parse date and year in Track::setDate to fix.

PR #295, opened 2024-09-26
Add $rand2() scripting function

lstrsrt:
This patch adds a `$rand2(min,max)` function, which is an extension to `$rand()` to allow custom bounds.

Foobar does not have this function, probably because it can technically be replicated with `$add(%min%,$mod($rand(),%max%))`. However, having another function for this is better because 1) it's more readable and 2) [mod causes skewed results.](https://stackoverflow.com/questions/10984974/why-do-people-say-there-is-modulo-bias-when-using-a-random-number-generator)

I also fixed a potential overflow in `Utils::randomNumber()`. (sidenote, this function could be called instead of using \<random\>, after changing it to return uint32_t... I don't know if Qt's implementation is any better though?)

### DIFF
--- a/src/core/scripting/functions/mathfuncs.cpp
+++ b/src/core/scripting/functions/mathfuncs.cpp
@@ -21,6 +21,15 @@
 
 #include <random>
 
+namespace {
+    uint32_t getRandomNumber(uint32_t min, uint32_t max)
+    {
+        std::mt19937 gen{std::random_device{}()};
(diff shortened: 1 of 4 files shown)


ludouzi:
Thanks, looks good.

> I don't know if Qt's implementation is any better though

Looking at the Qt source, it's implementation is very similar so it shouldn't matter either way. My preference now is to default to the standard library for most things unless the Qt implementation offers extra functionality we need.

ludouzi (approved):

PR #304, opened 2024-10-03
[i18n] Update zh_CN translations

CarterLi:
The only string left is `Elide text`. Don't really know what it means.

I found weblate exists after finishing all of these changes in an XML code editor. Boomer


ludouzi:
Many thanks for this! I was going to update this translation myself at some point so you've saved me a lot of work.

`Elide text` crops the text if it exceeds a certain length. What is really does it make the horizontal scrollbar visible, so I'll change the name of the setting so it better explains what it does.

And yeah, it would be best to use Weblate for any updates going forward as it's easier to handle merge conflicts if the translation sources are updated.

ludouzi (approved):

PR #310, opened 2024-10-03
Fix $iflonger behaviour

lstrsrt:
Previously, `$iflonger` compared the length of two strings, but according to [this reference](https://wiki.hydrogenaud.io/index.php?title=Foobar2000:Title_Formatting_Reference), `$iflonger` instead takes a numerical length as the second argument. `$longer` already covers comparing the lengths of two strings.


ludouzi:
Nice. Thanks for noticing that.

PR #328, opened 2024-10-13
Improve volume tooltip positioning

lstrsrt:
This patch fixes the volume control tooltip being partly hidden by the slider popup rect
when the tooltip is displayed on the left. The distance between the popup rect and the tooltip should now be equal no matter if the tooltip is on the left or right.


ludouzi:
I think this will need further adjustment. 

It works fine under Breeze, but using a style with narrow windows causes the tooltip to be displayed too far left:

image

I'll see if there's anything we can use from the style.

Issue #416, opened 2024-11-27
[FeatReq] %filesize_natural%


RedSnt:
This probably isn't highest on the list of things to focus on, and I'm grateful for %filesize% existing, it's just that using %filesize% as the dummy that I am have been a struggle. 
Made this custom column back in 0.7.1:

`$ifgreater(%filesize%,$mul($mul(1024,1024),1024),$div(%filesize%,$mul($mul(1024,1024),1024))" gb",$ifgreater(%filesize%,$mul(1024,1024),$div(%filesize%,$mul(1024,1024))" mb",$ifgreater(%filesize%,1024,$div(%filesize%,1024)" kb",%filesize%" bytes")))`

As you can see, not exactly graceful, but it works - would be easier with %filesize_natural% though.

Sidenote: I'm wondering how I trim the amount of decimals to 1 or 2 from the result I get: 

![image](https://github.com/user-attachments/assets/f4b64541-436d-482e-8175-7c2bb9369ad4)

Multiply by 1000, cut 2 from the right, then divide by by thousand again?

RedSnt:
It's been a little while, and because there hasn't been an official release in 8 months, I gave one of the nightly builds a go, and can confirm that %filesize_natural% works like a charm. Thanks a bunch.

PR #417, opened 2024-11-27
Add %filesize_natural%

lstrsrt:
Should address #416. A `$round` function seems like a reasonable idea too but I'll keep this PR to one new thing (also this current implementation already rounds down to the first decimal digit... but we could let `formatFileSize` take an additional argument for that).


ludouzi (approved):

ludouzi:
Thanks for handling that request.

Yeah, it'd probably be best to pass an additional arg to `formatFileSize` for control of rounding.

PR #434, opened 2024-12-27
[core] Prioritize "Playback Follows Cursor" over repeat and shuffle

### LINKED ISSUE #428: "Playback follows cursor" only works when skipping
Foobar2000 behavior: clicking on (selecting) a track with its equivalent of this option will "queue" it to play next whatever the shuffle/repeat setting is.

Fooyin behavior: clicking on (selecting) a track with "Playback follows cursor" on will "queue" it, but only if the "next" button is pressed. The next song that plays after the current song is finished is entirely dictated by the shuffle/repeat setting.

This seems counterintuitive, whether or not this app is meant to be a clone of foobar2000. Hopefully it's not a hard fix.

lstrsrt:
Fixes #428

### DIFF
--- a/src/core/playlist/playlist.cpp
+++ b/src/core/playlist/playlist.cpp
@@ -398,7 +398,7 @@ int PlaylistPrivate::getNextIndex(int delta, Playlist::PlayModes mode, bool only
     }
 
     if(m_nextTrackIndex >= 0) {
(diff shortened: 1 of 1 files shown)


ludouzi (approved):
Thanks!

PR #435, opened 2024-12-27
[core,gui] Properly update playlist on scheduled change

lstrsrt:
Another oneliner (+ typo fix), ran into this while doing #434. This fixes the icon in the tab bar not being updated when switching to a scheduled playlist (reproduced by enabling "Playback Follows Cursor" and switching to a track in a different playlist.)


ludouzi (approved):
Looks good. Thanks!

PR #436, opened 2024-12-27
[core] Engine: Fix seek interrupting fade-in

lstrsrt:
Fading in and then seeking during the fade causes the fade to be interrupted and the output volume to stay at whatever it was just before the seek. This is caused by `AudioPlaybackEngine::seek` calling `resetWorkers` which calls `AudioRenderer::resetFade`. Fix this by adding an argument that lets us skip the call.

In addition, call `resetFade` in `pause` to fix a bug where a long fade-in interrupted by an instant pause would not cause the fade timer to be stopped, meaning that another fade-in running before the previous timer was completed would cause the fade direction to flip and trigger a mute (introduced by #258).

### DIFF
--- a/src/core/engine/audioplaybackengine.cpp
+++ b/src/core/engine/audioplaybackengine.cpp
@@ -434,7 +434,7 @@ void AudioPlaybackEngine::seek(uint64_t pos)
         return;
     }
 
-    resetWorkers();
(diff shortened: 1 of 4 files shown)


ludouzi (approved):
Thanks. Good catch.

PR #442, opened 2025-01-06
[core] Scripting: Add $strstr and friends

### LINKED ISSUE #440: Feature request: scripting functions to return position of first/last occurrence of substr
I'm working on a grouping script, and it requires splitting a string at the occurrence of a substring. Example below:

Returning the substring that occurs before the first occurrence of "bar":
foobarbazqux -> foo
foobazbarqux -> foobaz
foobarquxbar -> foo

foobar2k includes a $strstr(str1,str2) function, which returns the position in str1 of the first occurrence of str2. Using this function, one could combine with the $left() function to produce the desired behavior

$left("foobarbazqux",$sub($strstr("foobarbazqux","bar),1)) -> "foo"

While foobar2k only provides a search for the first occurrence of a substring, I could also see the value of a search for the last occurrence of a substring as well, for example a $strstrright(str1,str2), with the following behavior:

$strstrright("foobarbazqux","bar") -> 4
$strstrright("foobazbarqux","bar") -> 7
$strstrright("foobarquxbar","bar") -> 10

lstrsrt:
Fixes #440

The str(i)strlast naming is just a suggestion, str(i)strright or rstr(i)str also sound okay to me -- feel free to change.

Usage examples:
`$strstr(testing,t)` -> 0
`$strstr(testing,t,1)` -> 3
`$strstr(testing,T,1)` -> {} (nothing)
`$stristrlast(testing,T)` -> 3

In addition (these are all small fixes so I included them in this PR):
* Add OOB check to `$slice` (prevents crash when n matches or exceeds string length)
* Restore `$insert(str,n)` functionality (broken since e14d900)
* Turn `elide_end` and `elide_mid` C++ names into camelCase


ludouzi (approved):

ludouzi:
Thanks for those fixes!

`$stristrlast` sounds good to me, so we can keep it as is.

PR #458, opened 2025-02-02
[core] Add support of audio/x-flac mime type

xordos:
This change is add support for audio/x-flac mime type which apparently used in Mac OS.  The detailed discussion is here:  https://github.com/fooyin/fooyin/discussions/438

I didn't wrap the change for Mac only as I feel this maybe useful for other distro as well,  and also inline with other mime type coding style.

I know Mac OS probably is not in socpe of this project, so this PR is 'nice to have' change, but as shown in above discussion, it actually worked pretty well in MacOS.

Thanks

### DIFF
--- a/src/core/engine/taglibparser.cpp
+++ b/src/core/engine/taglibparser.cpp
@@ -579,7 +579,7 @@ QString codecForMime(const QString& mimeType)
     if(mimeType == "audio/x-wavpack"_L1) {
         return u"WavPack"_s;
     }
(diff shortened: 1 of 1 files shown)


ludouzi (approved):

ludouzi:
Thanks!

There are plans for official macOS support in the future, so this will be useful for people who want to try using it before then.

xordos:
Ah, this is great news!  And thanks again for the hard work.  This player is *the* best music player I found so far.

PR #468, opened 2025-02-17
Fix #453 bugs with 'Playback follows cursor'

### LINKED ISSUE #453: "Playback follows cursor" still behaves strangely (latest)
After fixing https://github.com/fooyin/fooyin/issues/428, some residual bugs remain. 

- Double-clicking to start a song may cause the song to play twice in a row before moving on. 
- Sometimes, double-clicking any given song will restart the currently playing song instead, with the only remedy being to skip (usually restarting the same song again, but allowing the song I attempted to double-click before to play upon double-clicking it again) or double-click a different song. 

Both behaviors seem to happen randomly, with like a 70% chance of happening.

**System:**
 - OS: Arch Linux
 - fooyin version: fooyin-git 0.8.1.r152.g9db01ce-1

DaforLynx:
This fixes all the problems I found in https://github.com/fooyin/fooyin/issues/453, but I don't know for certain whether it introduces new bugs in other users' use cases.


DaforLynx:
Never mind...it only, somehow, makes them more inconsistent. In my initial testing the bugs did not appear, then I restarted the program and they came back. Don't merge yet.

DaforLynx:
Ok, I re-re-built and now it works again. It could be shenanigans involving the existing installation of fooyin on my system.

DaforLynx:
It does reset the next track when playing from stop or pause, but in my personal case that is fine.

DaforLynx:
Any idea when this can be reviewed?

ludouzi:
I'll look into it  in a bit. Just wanted to get a bugfix release out first to fix a few major bugs.

ludouzi:
Thanks for the attempted fix!

A key issue with the current behavior (even without this fix) is that selecting a track in a stopped state and then starting playback will not play that track. I also don't like tying the selected track to individual playlists anymore, as there can only ever be 1 selected track to play.

With that in mind, I think it'll make more sense to move that behavior to `PlayerController`, especially since it also handles the playback queue.

ludouzi:
https://github.com/fooyin/fooyin/commit/62a2c1df915093424365831e84fb37e8e61c5ae5 moves most of the logic to `PlayerController` so this is no longer necessary.

PR #479, opened 2025-03-06
[core,gui] Add "Follow playback queue"

### LINKED ISSUE #459: [Request] Option to Control Playback After Queued Song
Lets say this is the playlist:
Song 1
Song 2
Song 3
Song 4

If Song 3 is playing and Song 1 is added to the queue, the sequence is: Song 3 -> Song 1 (queued) -> Song 4. 

What I would except is Song 3 -> Song 1 (queued) -> Song 2, and so on. 

Pretty much every other music player that I've used works the latter way, so the current implementation in fooyin feels very counterintuitive. 

I understand that people could have differing opinions on this, so I suggest some sort of toggle for this in settings.

lstrsrt:
Fixes #459


ludouzi (approved):
Thanks! All working fine.

PR #480, opened 2025-03-07
Add CI for macOS

CarterLi:
This PR adds CI for both amd64 and arm64. The built artifacts (arm64 version) work for me in my MacBook M1 Pro

image


cmake/FooyinMacros.cmake
```diff
@@ -1,11 +1,38 @@
 function(fooyin_set_rpath target prefix)
     get_filename_component(path "${CMAKE_INSTALL_PREFIX}/${prefix}" ABSOLUTE)
     file(RELATIVE_PATH relative ${path} "${CMAKE_INSTALL_PREFIX}/${LIB_INSTALL_DIR}")
-    set(rpath "\$ORIGIN/${relative}")
```
CarterLi, line 4: Not sure about this. `${relative}` was empty when building in CI, which made fooyin binary unable to find dylibs

PR #505, opened 2025-04-06
Fix M3U extension (was M3)

kode54:
This looks like there was some sort of search and replace for the sequence `u"`


kode54:
Whee, looks like the Arch CI failed due to libvgm's then-current commit hitting CMake 4.0. libvgm has fixed this upstream, so I guess I'll bump that in another PR.

ludouzi (approved):
Yeah, that was probably caused by the regex replace of `QStringLiteral`. 

Thanks for finding that!

PR #515, opened 2025-04-24
Decouple restoring playback state from active playlist state

luca020400:
It's convinient to have the player automatically
restore the previous playlist state but now the playback.

### DIFF
--- a/src/core/application.cpp
+++ b/src/core/application.cpp
@@ -99,6 +99,9 @@ public:
     void startSaveTimer();
     void exportAllPlaylists();
 
+    void saveActivePlaylistState() const;
+    void loadActivePlaylistState() const;
+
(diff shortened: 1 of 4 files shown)


luca020400:
Just wanted to get your opinion on this, probably the "save playback state" should be a dependency on save active playlist?
This also breaks users on update since they won't have the playlist anymore.

anyways, please take a look! ty

luca020400:
@ludouzi any opinion on that?
I really like preserving my playlist state, but I don't want to auto-play on open

If you're okay with it I'll happily rebase

ludouzi:
Yeah, I think this will be a useful option. 

Please rebase it so it can be merged into the next bugfix release.

luca020400:
Looks good te me, works fine

PR #516, opened 2025-04-25
Align click actions of library tree and browser page

luca020400:
Library tree now has a better name for the 'Play' action.
Browser page gains 'None' action for double-click.
Browser page gains 'Add to playback queue' action for middle click.


ludouzi (approved):
Good changes! 

The main reason the library tree play action was named just `Play` was to avoid confusion with the `None` action which can also expand/collapse. Though that shouldn't be a problem for most users and it does make it consistent with other widgets.

PR #524, opened 2025-05-17
Fix tests

ypsvlq:
The test data library is ignored if  `--as-needed` is enabled (which it is by default on some systems).


ludouzi (approved):
Thanks!

Issue #543, opened 2025-06-02
playlist auto-export ignores path type setting and only stores absolute paths and never relative paths


0x4ce66f11:
**Describe the bug**
_Settings > Playlist > Saving > Path Type_
When _"Path Type"_ is set to _"Relative"_, manually saving a playlist works as expected and the playlist file contains relative paths.
However the playlist auto-export feature automatically saves playlists files containing absolute paths, ignoring the _"Path Type"_ setting.

**To Reproduce**
1) Navigate to _"Settings > Playlist > Saving > Path Type"_
2) Change the _"Path Type"_ setting to _"Relative"_
3) Navigate to _"Settings > Playlist > Auto-export"_
4) Check _"Auto-export"_ to enable it
5) Select a directory where the playlists are automatically saved to
6) Apply settings
7) Create a new playlist
8) Wait for it to be auto-exported
9) Open the resulting playlist file in a text editor to inspect the result

An example directory structure:
* `Music/Lossless/` <- contains music files
* `Music/Lossy/` <- contains music files
* `Music/FooyinPlaylists/` <- contains auto-exported playlists

**Expected behaviour**
The auto-exported playlist files should have the same path-type as manually exported playlist files.

**Screenshots**
![Image](https://github.com/user-attachments/assets/4f03ff24-e12e-46cd-a18d-36a0c7f32616)

**System:**
 - OS: Linux openSUSE Tumbleweed-Slowroll 20250501
 - fooyin version: 0.8.1 (flatpak)

evargh:
Function `ApplicationPrivate::exportAllPlaylists` doesn't actually read the setting you set there. It's initialized with a default value and then is passed to the `savePlaylist` function used by both manual and automatic exporting.

https://github.com/fooyin/fooyin/blob/78b9ba1528eea485c3dde2838984b9f0a02a64fb/src/core/application.cpp#L284

ludouzi:
Fixed in https://github.com/fooyin/fooyin/commit/4400354efd2d3fd649c02965e503301b1e3cd159.

PR #545, opened 2025-06-05
[core] Fix playlist auto-export not obeying settings

evargh:
Tiny fix for #543. This ultimately means that `ApplicationPrivate::exportAllPlaylists` and `SavePlaylistsDialogue::accept` have very similar logic. I wonder if putting this function into `utils` would be a better move?


ludouzi (approved):

ludouzi:
Thanks! 

There is similar logic between those two functions, but enough differences to where it's not an issue for now.

PR #565, opened 2025-06-20
[plugins] Scrobble Filtering Feature

evargh:
This commit implements the feature used in the Foobar2000 scrobbling plugin, where queries can be used to filter which tracks are scrobbled.

This was a feature I liked in Foobar2000's plugin, since I used it to listen to lectures and podcasts alongside music and didn't want those scrobbled. It's your call on whether this feature is in scope for Fooyin.


ludouzi (approved):
Thanks! That's a useful feature, so it definitely belongs in fooyin.

PR #570, opened 2025-06-23
[plugin] fix messy code when selecting output sound device

### LINKED ISSUE #568: Messy code when selecting output sound device
**Describe the bug**

See the screenshot. Happens when the name of my sound device contains CJK characters. Reproducable in both SDL2 and Pipewire mode.

**To Reproduce**

Settings->Playback->Output

**Expected behaviour**

CJK characters are displayed correctly

**Screenshots**

Image

**System:**
 - OS: Arch Linux
 - fooyin version: git head

**Additional context**

CarterLi:
IMO `QString::fromLatin1` should never be used. There are a lot places that still uses `fromLatin` that should be double checked.

Fixes #568

### DIFF
--- a/src/plugins/alsa/alsaoutput.cpp
+++ b/src/plugins/alsa/alsaoutput.cpp
@@ -123,11 +123,11 @@ void getPcmDevices(Fooyin::OutputDevices& devices)
         }
 
         if(strcmp(name.str, "default") == 0) {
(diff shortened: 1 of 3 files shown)


ludouzi (approved):

ludouzi:
Thanks!

`QString::fromLatin1` is fine for compile-time strings and cases where the encoding is known. Still, I'll go over the remaining usages to identify any similar issues.

PR #597, opened 2025-07-26
[core] Engine: Fix playback restarting when stopping with fade out

lstrsrt:
Back with another fade fix! This time for stopping instead of seeking.

Pausing and then stopping during the fadeout breaks playback state so that playback continues after the fadeout completes, but the UI looks like playback was stopped. This also breaks shutdown. Fix by making `stopEngine` be called if we're being stopped during FadingOut state.

### DIFF
--- a/src/core/engine/audioplaybackengine.cpp
+++ b/src/core/engine/audioplaybackengine.cpp
@@ -1,6 +1,7 @@
 /*
  * Fooyin
  * Copyright © 2023, Luke Taylor <LukeT1@proton.me>
+ * Copyright © 2025, Gustav Oechler <gustavoechler@gmail.com>
  *
  * Fooyin is free software: you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


lstrsrt:
Btw given that I've made a couple of commits to this file, can I add myself to the copyright header? Also the year on your line needs to be updated.

ludouzi:
Thanks! Glad to see you back.

> Btw given that I've made a couple of commits to this file, can I add myself to the copyright header? Also the year on your line needs to be updated.

Sure thing. I don't actually update the year in the copyright header; I like to keep it as the year the file was created. Plus it avoids unnecessary commits updating every file each year.

lstrsrt:
Done.

>  I don't actually update the year in the copyright header; I like to keep it as the year the file was created.

Sure, kept as is to avoid inconsistency

PR #643, opened 2025-08-30
[OpenMPT] Added loop count option to OpenMPT settings

kquote03:
Hello,
Added loop count option to openmpt, just like the other tracker modules (vgm input and game music emu) because it was driving me mad
Please let me know if there are any issues, it is my first time pushing c++ code lmfao
thanks


ludouzi (approved):
Thanks!

No issues which I can see.

PR #668, opened 2025-09-13
[gme,ompt,libvgm] Fix infinite looping

### LINKED ISSUE #667: Repeat Track does not make any of the looping formats loop indefinitely
**Describe the bug**
Looping formats do not loop indefinitely when the player is set to Repeat Track. They proceed to fade out and terminate, then restart.

**To Reproduce**
1. Set player to Repeat Track
2. Queue some GME, MOD, or PSF files.
3. Set them to play.

**Expected behaviour**
They should play past their duration without fading, and continue playing until stopped.

**System:**
 - OS: Arch Linux
 - fooyin version: 0.9.1.r46.gcaeb3da-1

**Additional context**
I just implemented support in my plugins, too.

kode54:
Fixes 42e838cc958e4466b76994b694474409a9bf43a4
Fixes #667

Also fixes preferences dialogs for looping settings, including OpenMPT, where the loop count is additional loops, not counting the initial playthrough.

Also, OpenMPT needs the renderer parameters to be set after selecting the subsong, not before.

### DIFF
--- a/src/plugins/gme/gmeinput.cpp
+++ b/src/plugins/gme/gmeinput.cpp
@@ -130,7 +130,7 @@ Track GmeDecoder::changedTrack() const
 
 std::optional<AudioFormat> GmeDecoder::init(const AudioSource& source, const Track& track, DecoderOptions options)
 {
-    m_options = options;
+    m_repeatTrack = !(options & NoInfiniteLooping) && isRepeatingTrack();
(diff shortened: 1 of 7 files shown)


ludouzi (approved):
Perfect timing, I was just about to work on this too.

Thanks!

PR #703, opened 2025-09-30
Add initial Windows build support

ludouzi:
Adds initial support for building on Windows using vcpkg. CI builds an installer and a ZIP package.

Documentation to follow.


CarterLi:
Hello, the icon `src\data\icons\fooyin.ico` is blurred on my HiDPI monitor.

{AD71C924-340F-4B18-BA00-7090FCF3B03F}

Better to have 512x512 icon bundled in too.

ludouzi:
> Hello, the icon src\data\icons\fooyin.ico is blurred on my HiDPI monitor.

Yeah, that was caused by only rasterizing once from the svg. The ico is now created using the pngs which should fix the issue, but I also added 512x512 just in case.

PR #752, opened 2025-10-30
[gui] Fix PropertiesDialog not auto-defaulting to Ok button

lstrsrt:
Opening the properties dialog, changing the metadata and pressing Enter closes the dialog but discards all changes, because the dialog auto-defaults to the Cancel button. I don't think this was intended, given that `setDefault(true)` is called on Ok. The missing step is to also call `setAutoDefault(false)` on Cancel.


ludouzi (approved):

ludouzi:
Yeah, that wasn't intended. Thanks!

PR #760, opened 2025-11-08
CLI - add seek commands

tmtom:
Added seek commands which could be used for example for global shortcuts. Seek amount is in ms.
E.g.
`fooyin --seek-forward 5000`
will seek/skip 5s forwards. If preferred I can change to a single command and using negative seek time for backward and positive for forward.


lstrsrt:
The platform ifdefs can be avoided with `std::stol`

ludouzi (approved):
Looks good. Thanks!

PR #788, opened 2025-12-12
[feature]: Drag playlist items to external applications

donutinit:
This change adds support for dragging tracks from Fooyin’s playlist UI to external applications (system-wide drag-and-drop). The goal is to make it possible to select items in the playlist and drag them out of Fooyin into other apps (e.g., file managers, editors, chat apps that accept dropped files/URIs, etc.).

### Notes / limitations

Tested only on Fedora 43 Gnome Wayland.

I cannot guarantee it works on other distributions or platforms.

This is the first application I’ve compiled and contributed changes to, so I cannot guarantee the implementation is perfect or production-grade.

Edit: also tested wiht Cinnamon on XORG and it works flawlessly

### DIFF
--- a/include/utils/helpers.h
+++ b/include/utils/helpers.h
@@ -72,10 +72,14 @@ Cntr filter(const Cntr& container, Pred pred)
 template <typename Cntr>
 Cntr sortByIndexes(const Cntr& target, const std::vector<int>& indexes)
 {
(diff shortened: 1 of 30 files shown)


evargh:
`CMakeCache.txt` is still in the PR

Issue #694, opened 2025-09-25
Feature request: Adding support and/or chaptering for audiobook m4b and other container type.


mwdv:
I prefer my audio file in one place including audiobook. like in foobar.

tmtom:
Also would love this feature - in my case multitrack m4a which is basically the same thing (Foobar supports it). Thanks

PR #790, opened 2025-12-14
[draft] ffmpeg chapters

kode54:
Here is a draft, showing it's possible to read chapters. With some work, one could even use FFmpeg to write tags by altering the tag dictionaries, then transmuxing to a new file, closing, then moving the new file into place.

There are some workarounds added for now, which allow the tag reader to take priority, or else TagLib will prevent the chapters from being seen.

I'm open to any further work this might need, but perhaps you could take a stab at redoing parts of it.

### DIFF
--- a/src/core/engine/input/ffmpeg/ffmpeginput.cpp
+++ b/src/core/engine/input/ffmpeg/ffmpeginput.cpp
@@ -186,7 +186,7 @@ QString convertString(const char* str)
     return QString::fromUtf8(str);
 };
 
-void parseTag(Fooyin::Track& track, AVDictionaryEntry* tag)
+void parseTag(Fooyin::Track& track, const AVDictionaryEntry* tag, bool isTrack, int nbChapters)
 {
     if(strcasecmp(tag->key, "album") == 0) {
         track.setAlbum(convertString(tag->value));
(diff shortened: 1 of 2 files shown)


kode54:
Gonna run into a lot of CI failures due to Qt 6.10.1.

kode54:
Hi, clang-format 21.1.6 produces different results for the repo included config file than the version your CI uses.

Edit: Looks like "Empty" is not actually a valid setting for any clang-format options, and it will just use whatever that version's default is.

tmtom:
Thanks for you work, was able to rebase on master and cleanly build on Ubuntu 24.04.
I believe this would resolve https://github.com/fooyin/fooyin/issues/694, right?
Did quick check on a m4a with chapters but did not worked for me. Tried m4a create by https://github.com/yt-dlp/yt-dlp
using `-f bestaudio --format-sort acodec:m4a --embed-thumbnail --embed-metadata --add-metadata` and downloaindg random YT video with chapters (this creates m4a with chapters and cover art) but still shows as a single track in fooyin.

kode54:
Unfortunately, I don't know how it caches track info. It has this annoying behavior where any other Reader class will take it over and present a single track. I tried to override it so it places the FFmpeg Reader as first.

Unfortunately, this also conflicts with tag writing for the supported formats. An FFmpeg tag writer will need to be added to this.

Edit: Oh, you can reorder them yourself. You'll need to drag the FFmpeg tag reader above the TagLib one.

kode54:
Sorry for the noise, it should be good for testing now.

ludouzi:
Thanks for working on this.

I’ll take another look at #882 first (and make the suggested improvements), then give this a test.

kode54:
This could be reworked to use similar interfaces to the CUE handling. The only major problem at this point is the tag reading vs writing, since it needs the tag reading to get the album/track titles and ReplayGain correct for FFmpeg, and there is no writing implementation, since FFmpeg requires remuxing an entire file to a new file to rewrite metadata.

PR #801, opened 2026-01-02
[notify] Add track change desktop notifications

### LINKED ISSUE #347: Desktop notifications
I just discovered this wonderful player so maybe I just didn't find the option yet but would be nice to have desktop notifications on song change.

A small notification with album art, title, album and artist would be enough.

ripdog:
Summary
- Adds a new notify plugin to show desktop notifications on track changes (Linux / org.freedesktop.Notifications via D-Bus).

Features
- Configurable notification title/body using Fooyin's script/tag formatting.
- Optional album art via CoverProvider + D-Bus image-data hint.
- Settings page to enable/disable and configure fields/timeout/artwork.

Tested and working on Arch Linux, KDE Plasma, Wayland.

It's Linux-only, and sends via D-Bus. This is because there doesn't seem to be a Qt method for sending desktop notifications. QSystemTrayIcon exists, but requires displaying a tray icon before notifying, and scales the image down very small. Basically, if we want to notify on Win/Mac, native code for each is ideal.

Full disclosure: This is AI generated code. I have reviewed and tested personally, and am ready and willing to make changes as necessary.

Closes #347

### DIFF
--- a/src/plugins/notify/CMakeLists.txt
+++ b/src/plugins/notify/CMakeLists.txt
@@ -1,12 +1,25 @@
+find_package(Qt6 COMPONENTS DBus QUIET)
+
+if(NOT Qt6DBus_FOUND)
+    message(STATUS "Qt6::DBus not found; skipping notify plugin.")
+    return()
+endif()
+
+set(SOURCES
+    notifyplugin.cpp
+    notifyplugin.h
+    settings/notifypage.cpp
+    settings/notifypage.h
+    settings/notifysettings.cpp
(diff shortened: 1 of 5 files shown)


ludouzi:
Thanks for the PR!

I had to make a few changes as the plugins `CMakeLists.txt` has changed significantly. I also avoided blocking the main thread by using `QDBusPendingCallWatcher`, and switched to a generated D-Bus interface XML for `org.freedesktop.Notifications`, rather than using raw calls.

PR #810, opened 2026-01-13
Add sort option to Playlist Organizer context menu

warhuhn:
I've added a "sort playlists" and "sort playlists in group" options to the Context menu in the Playlist Organiser.

They sort all playlists, or just playlists inside a group alphabetically, ascending.
I later thought how many people would sort them descending, which would be few, and adding more options would just clutter the menu.

I am pretty sure this isn't ideal, and i haven't done anything with C++ in a long time, but it works pretty well for my purpose.


ludouzi:
Thanks!

I made a few changes/fixes, including taking a snapshot of the expanded/view state before and after sorting, as the saved state is only saved on shutdown. I also tightened the sort actions so they don't sort groups as well.

PR #841, opened 2026-02-11
Add playback setting to load directory when opening file from file browser

DocMAX:
Adds a new playback setting 'When opening file via file browser open dir to playlist and play opened file' which:

1. When opening a single music file from the file browser
2. Loads all files from the same directory into a playlist
3. Plays the selected file
4. Ensures the playlist scrolls to the playing track

I like this behavior from the old Windows player "Billy MP3"


ludouzi:
Thanks for the PR!

I made some changes to adapt it to newest master, and moved the setting to the `Shell Integration` page where it's a better fit.

ludouzi:
Additional changes are needed to have the playlist scroll to the current track on open, but I'll handle that later, as it'll be more involved given the recent changes to playlist loading.

ludouzi (approved):

DocMAX:
Thanks a lot for accepting my idea

PR #850, opened 2026-02-16
Added settings to show/hide hidden files/directories and symlinks

Birbwell:
Both settings are disabled by default, and can be found in Settings>Widgets>Directory Browser.

Added these features since I like to form my music directory from multiple symlinks to other music libraries (like the Steam Music folder).

Right now the settings occupy the widget with the title "Browser Mode," but under a separate title "Browser Filters." I did this rather than being creating their own widget or including it with the Display Options, since they seemed more similar to the Browser Mode options than the other Display Options. Creating a new widget causes the default window size to be too small, meaning you have to scroll, and I'd rather avoid that.

The existing default filters have been moved into their own private variable within the directory browser class, then gets bitwise-or-assigned with the "NoSymLinks" and "Hidden" flags depending on what the settings are set to. On-update, the filters get re-applied to the model.

### DIFF
--- a/src/gui/dirbrowser/dirbrowser.cpp
+++ b/src/gui/dirbrowser/dirbrowser.cpp
@@ -247,8 +247,8 @@ DirBrowserPrivate::DirBrowserPrivate(DirBrowser* self, const QStringList& suppor
     , m_dirTree{new DirTree(m_self)}
     , m_model{new QFileSystemModel(m_self)}
     , m_proxyModel{new DirProxyModel(m_mode == DirBrowser::Mode::List, m_self)}
-    , m_showSymLinks{static_cast<bool>(m_settings->value<Settings::Gui::Internal::DirBrowserShowSymLinks>())}
-    , m_showHidden{static_cast<bool>(m_settings->value<Settings::Gui::Internal::DirBrowserShowHidden>())}
+    , m_showSymLinks{m_settings->value<Settings::Gui::Internal::DirBrowserShowSymLinks>()}
+    , m_showHidden{m_settings->value<Settings::Gui::Internal::DirBrowserShowHidden>()}
     , m_doubleClickAction{static_cast<TrackAction>(m_settings->value<Settings::Gui::Internal::DirBrowserDoubleClick>())}
     , m_middleClickAction{static_cast<TrackAction>(m_settings->value<Settings::Gui::Internal::DirBrowserMiddleClick>())}
     , m_context{new WidgetContext(m_self, Context{Constants::Context::DirBrowser}, m_self)}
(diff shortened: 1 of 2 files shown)


Birbwell:
The Arch Linux build test failed due to a pacman transaction error, but I can personally attest to it building and running on Arch as that is the distro I used to build/test it :)

Also, while I'm here, would it be possible to add pipewire-alsa to the AUR pkgbuild as a dependency (if you are the ones maintaining it)? It's required for the application to play music, but does not automatically install when installing via the AUR.

src/gui/dirbrowser/dirbrowser.cpp
```diff
@@ -239,6 +247,8 @@ DirBrowserPrivate::DirBrowserPrivate(DirBrowser* self, const QStringList& suppor
     , m_dirTree{new DirTree(m_self)}
     , m_model{new QFileSystemModel(m_self)}
     , m_proxyModel{new DirProxyModel(m_mode == DirBrowser::Mode::List, m_self)}
+    , m_showSymLinks{static_cast<bool>(m_settings->value<Settings::Gui::Internal::DirBrowserShowSymLinks>())}
+    , m_showHidden{static_cast<bool>(m_settings->value<Settings::Gui::Internal::DirBrowserShowHidden>())}
```
ludouzi: The casts aren't needed here.

ludouzi:
> Also, while I'm here, would it be possible to add pipewire-alsa to the AUR pkgbuild as a dependency (if you are the ones maintaining it)?

I'm the maintainer of the `fooyin-git` package, but `pipewire-alsa` is listed as an optional dependency of `pipewire`, which itself is an optional dependency of `fooyin`. It's best to keep it optional as not everyone may want/need pipewire.

Birbwell:
I've taken your recommendations into account. Just going to test the UI change locally before pushing here :)

Also, thank you for clarifying what the UI design philosophy! I originally gave the filter options their own QGroupBox, but decided against it because it made it so the page needed to scroll (even if slightly). I wasn't sure if that was ok, so I came up with something else that wouldnt cause scrolling, hoping I'd get feedback on what to do later lol.

And then it makes sense why you'd want to keep `pipewire-alsa` optional. I may look into adding better error reporting when it comes to alsa, then, because if you don't have anything alsa-related installed it'll just give a generic "Unable to open playback device" error. When I first started using `fooyin`, it took me a bit to figure out what was going on lol.

Birbwell:
If one of the checks fails due to a runner networking issue, would you prefer it to be left alone, or would you want the checks to be re-triggered by committing again?

ludouzi:
Thanks! I re-triggered the failing runner.

> if you don't have anything alsa-related installed it'll just give a generic "Unable to open playback device" error.

I'll have to look into that. The ALSA plugin won't actually get built and be included as an output option if the ALSA library isn't installed. That error could possibly have been caused by an incompatible audio format.

ludouzi (approved):

Birbwell:
Maybe. Idk, the audio plays fine once you install the `pipewire-alsa` package (on Arch, at least). Before that, no matter what you try to play just gives that error (in my experience, at least).

PR #858, opened 2026-02-21
Engine v2: new playback architecture

ludouzi:
# Engine v2: new playback architecture

## Why

The previous engine (`AudioPlaybackEngine` and `AudioRenderer`) was implemented as a two-threaded monolith. 
One thread handled both engine interaction and decoding, while the other performed gain processing, resampling, 
and output. 

Communication relied on QThread and signal-slot interactions, which introduced frequent race conditions 
and significantly limited extensibility. In particular, the design made features like DSPs and crossfading impractical to implement.

This change introduces a layered architecture that separates responsibilities into clearly defined, independently testable components. It enables both crossfading and a flexible DSP pipeline. Bit-perfect playback is still preserved when no DSP is active, ReplayGain is disabled, and internal volume is at 100%.

### Threading model

The engine is now split across more threads:

| Thread              | Responsibility                                           |
|---------------------|----------------------------------------------------------|
| **Main UI thread**  | Transport intent, playlist navigation, position display  |
| **Engine thread**   | Track lifecycle, decode control, transition coordination |
| **Audio thread**    | Mixing, DSP, output write, timing                        |
| **Prepare worker**  | Background preparation of the next track                 |
| **Analysis worker** | Post-master PCM analysis; level meter computation        |

### Decode pipeline

- `DecodingController` runs a timer-driven decode loop on the engine thread, feeding decoded PCM
  into a ring buffer in the stream's native format.
- `AudioStream` wraps a lock-free SPSC ring buffer and exposes a frame-by-frame read interface
  consumed by the audio thread. Multiple streams can be active simultaneously, enabling crossfade and gapless transitions.
- `NextTrackPrepareWorker` opens and prepares the next track in the background so its stream is
  ready before the current one ends.

### AudioMixer and per-stream processing

`AudioMixer` owns all active `AudioStream`s and produces a single mixed F64 buffer per audio cycle. For each stream, it applies a three-stage per-stream processing chain before accumulation:

1. **`TrackProcessor`s** (e.g. `ReplayGainProcessor`) - gain adjustments from track
   metadata, applied before any DSP.
2. **Per-track `DspChain`** - user-configurable DSP nodes that run on a single stream before it
   is mixed with others.
3. **Fade ramp** - amplitude envelope for crossfade and seek in/out.

The mixer now carries source-timeline metadata through per-track DSP buffering and returns it alongside audio (`ReadResult::primaryTimeline`). This timeline is used later to map written output back to source position.

Crossfade and gapless transitions are managed directly in the mixer. Pending streams are warmed before handoff to absorb DSP startup latency, and gapless transitions can splice the next stream into the same audio cycle when the outgoing stream ends mid-block. `OrphanStreamTracker` handles orphan stream cleanup once the fade completes.

### Master DspChain

After mixing, the coalesced F64 buffer passes through a **master `DspChain`** - a list of `DspNode`s that operate on the full mix. Both chains support format-changing nodes by propagating `outputFormat()` through the chain at `prepare()` time.

### OutputFader and AudioOutput

After the master chain, `OutputFader` applies a fade envelope used for pause, stop, and track change fades, with configurable curves (linear, sine, cosine, logarithmic). The final buffer is converted from F64 to the output device's native format (with optional dithering) and written to the `AudioOutput` backend (ALSA, PipeWire, SDL, etc.).

### AudioAnalysisBus

During render, post-master PCM is pushed into an `AudioAnalysisBus` via a ring buffer with a presentation timestamp. A dedicated worker thread drains the ring buffer, accumulates fixed 1024-frame hops, and computes `LevelFrame` data (per-channel peak and RMS). Other types of data will be added in the future.
Results are dispatched via callback to `AudioEngine`, which forwards them to the UI for visualisers.  Subscriptions are flag-controlled, so the analysis path is fully skipped when nothing is listening.

### Source-timeline position tracking

Previous position handling used a separate std::steady_clock running in parallel with the engine. Although started and stopped with playback, it was not synchronised to the actual audio stream. Position reporting is now derived from the source-timeline position of audio submitted to the output device, rather than from an independent parallel clock:

- Per-stream reads produce timeline segments (`startNs`, `frameDurationNs`, `streamId`) that are
  rebuilt into `ProcessingBufferList` chunks before master DSP.
- `AudioPipeline` stores timeline chunks in the pending output FIFO (including timeline epoch) and
  maps write offsets back to source positions when frames are actually written.
- Mapping handles unknown regions by using the last mapped boundary and a fallback source-time scale
  derived from observed master output rate.
- Epoch/stream mismatches and large regressions are rejected; the pipeline falls back to decode-head
  position until mapped data is trustworthy again.
- Delay publication is basis-aware: mapped source position subtracts output/device queue delay,
  while decode-head fallback includes DSP latency compensation.

### AudioClock and position publication

`AudioClock` maintains a linear anchor-based position extrapolator:

- `AudioEngine` periodically samples pipeline status (or on sync requests) and calls
  `AudioClock::applyPosition(sourcePositionMs, outputDelayMs, delayToSourceScale, ...)`.
- `AudioClock` anchors on discontinuities (seek/switch/generation change), then extrapolates
  smoothly between anchors using the reported playback-rate scale.
- Two timers drive publication: a sync request timer and a publish timer, so UI position stays
  smooth even between explicit pipeline updates.

### DSP plugin system

- `DspPlugin` and `DspNode` are the core interfaces.
- Plugins can implement `DspSettingsProvider` to provide UI for configuring DSP nodes.
- `DspChain` sequences nodes and accumulates total latency; format changes propagate through the
  chain at prepare time.
- `DspRegistry` maps node IDs to factories; third-party plugins register via the plugin API.
- Built-in nodes: `ResamplerDsp`, `SkipSilenceDsp`, `MonoToStereoDsp`, `ReverseStereoDsp`,
  `DownmixToStereoDsp`, `DownmixToMonoDsp`.

---

## Audio pipeline cycle

```
1. Drain command queue
2. AudioMixer::read(frames)
   ├─ For each active stream:
   │   ├─ Read PCM from AudioStream ring buffer
   │   ├─ TrackProcessors: ReplayGain gain
   │   ├─ Per-track DspChain
   │   └─ Fade ramp × accumulate into F64 mix buffer
3. Master DspChain::process(mix buffer)
4. OutputFader::apply(mix buffer)
5. Push post-master PCM → AudioAnalysisBus (with presentation timestamp)
6. Queue post-master chunks into pending output FIFO (format convert + timeline chunks)
7. Drain pending output FIFO → AudioOutput
8. Map written output offsets → source-timeline position (epoch-aware, mapped/fallback)
9. Update pipeline delay/position status
10. Engine publishes to AudioClock
```


magusjanus:
One issue with fooyin that has kept me away from using it is that there is a noticeable delay when pausing and playing. It doesn't make it unusable but it's a "every other player has solved this problem, why not this one?" This playback architecture rewrite branch has fixed that, thanks a lot for fixing it. This makes me a lot more interested in using fooyin.

However there is still a delay when manually changing to a new track, on both master and this branch. Hope that can inform further improvements to this rearchitecting.

ludouzi:
Thanks for testing and for the feedback. Glad to hear pause/play latency is improved on the rewrite branch.

For track changes, I can’t reproduce a noticeable delay on my setup (on both master and this branch), so this may be backend/config dependent. If you’re willing, it would be great if could you share:

  1. OS + audio backend (ALSA/PipeWire).
  2. Playback settings: buffer length, gapless.
  3. Whether the delay is with all files, or only archive/network files.

magusjanus:
I just tested both [2ce06a2](https://github.com/fooyin/fooyin/commit/2ce06a26719669fa1277c4b6b967baece0df9fb7) and the latest commit [f9a37e2](https://github.com/fooyin/fooyin/commit/f9a37e2bc336648ce7f761b1b91df471cab28f98) back-to-back, and I can confirm that the track changing delay has been fixed as of the latest commit!

My setup was:
1. CachyOS rolling, ALSA+PipeWire
2. Gapless playback, Buffer length 2000ms (default settings)
3. The delay was happening with local files on SSD.

ludouzi:
Great to hear!

PR #875, opened 2026-03-05
Adjust lyrics panel colors when theme changed

luyu-wu:
currently when changing theme (dark <-> light), the lyrics panel color (text, background, etc.) will not update.

- ive written some code that fixes this by not storing the colors as a static list. This doesn't break custom colors users may set.

- instead of using darken for out-of-focus text as the default (which only affect dark mode), I switch this to placeholder text (which achieves the same effect with colorscheme ambivalence).

I'm hoping to fix playlist tracks not updating color and icons as well, but maybe a later merge (they seem more complicated)

### DIFF
--- a/src/plugins/lyrics/lyricscolours.h
+++ b/src/plugins/lyrics/lyricscolours.h
@@ -38,26 +38,34 @@ struct Colours
         WordSynced,
         LineUnsynced
     };
-    
+
     QMap<Type, QColor> lyricsColours;
-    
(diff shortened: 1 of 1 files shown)


ludouzi:
Thanks for the fixes!

This will need a few follow up changes as resolved defaults are still saved to the config, and the lyrics text itself will also need refreshing.

But the latter is part of a bigger change that will be needed for RichText, which is now used in most stock fooyin widgets. I'll take a look into that.

PR #882, opened 2026-03-06
[core,plugins] Multiple formats per extension

kode54:
Support scanning all supported inputs in priority order.

Though maybe there's a better way of doing this?

### DIFF
--- a/include/core/engine/audioloader.h
+++ b/include/core/engine/audioloader.h
@@ -28,6 +28,40 @@ namespace Fooyin {
 class AudioLoaderPrivate;
 class Track;
 
+struct LoadedSource
+{
+    //! Non-owning view passed into decoder/reader backends.
+    AudioSource source;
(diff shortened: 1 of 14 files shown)


ludouzi:
Good idea.

Since callers only ever need a single usable decoder or reader, it would be cleaner to keep the iteration and initialisation inside `AudioLoader`, rather than returning a list of candidates and having each call site try them itself. That avoids duplicating the fallback logic in multiple places. In that case, returning a small result struct containing the initialised decoder/reader along with the associated file/device ownership would make sense.

The existing methods could still be kept in case a caller ever does need the full lists.

ludouzi:
Implemented along those lines.

The decoder/reader selection and initialisation now live in `AudioLoader`, with small result structs for the opened source plus the selected decoder/reader. The methods which return lists of all compatible readers/decoders are still there and used for the new load methods.

PR #883, opened 2026-03-06
[gme] Fix playback startup

kode54:
Playback repeats itself within about 3 seconds of startup if start() is allowed to restart the track.


ludouzi:
Thanks. This only became an issue with the engine rewrite: we now prefill in a background thread, then adopt that prepared decoder later, so the same decoder can see start() twice without a stop(). 

I'll get around to guarding that at the engine level at some point.

Issue #843, opened 2026-02-13
Per-layout widget settings or widgets with custom settings


robin-mu:
Currently, widget appearance settings are global and do not depend on the layout. A layout only saves settings from the "Interface" page.

Examples:
-  I can have layouts with different lyrics fonts (since this is a setting from the "Interface" page), but the current lyrics line font can not be different across layouts (since it is a setting from the "Lyrics" page).
- It is not possible to have two layouts with a different status widget texts (since it is a setting from the "Status Widget" page).

To solve this, there could either be a setting in the "Export Layout" dialogue to also save widget appearance settings.
Or it could be possible for widgets to override the global widget appearance settings and specify their own.

Also, this is related to #410

ludouzi:
Agreed. It would make more sense for most of the settings under the `Widgets` category to apply on a per-widget basis, so they are saved automatically with layout files.

The best approach will likely be to add a right-click option for widgets with configurable settings that opens a config dialog. I just need to decide whether to keep the global settings pages as well, since they would duplicate the same options which could become confusing. Maybe a `Save as default` button could be added to each config dialog.

PR #885, opened 2026-03-07
Add per-widget configuration

ludouzi:
This introduces a shared per-widget configuration for most layout widgets.

Each widget now has a `Configure...` action in its instance context menu, opening a widget-specific config dialog. Widget instance state is now serialised into the layout file, so configuration is restored per instance when a layout is loaded.

As part of this change, widget options that were previously only available through global settings pages have been moved into per-widget configuration where appropriate. The corresponding global settings pages have been removed, except in cases where the underlying state remains genuinely global or shared across multiple widget instances.

Default configuration for newly created widget instances remains persisted through the existing settings store and keys. The config dialogs provide a `Save as default` action, which writes the current dialog state back to those stored defaults without affecting already-configured widget instances.

In addition to the above, all settings which did not make use of `subscribe` have been demoted to unregistered settings, after which they must use `fileValue` and `fileSet` instead.  

Addresses #843

### DIFF
--- a/src/gui/dirbrowser/dirbrowser.cpp
+++ b/src/gui/dirbrowser/dirbrowser.cpp
@@ -61,7 +61,6 @@
 using namespace Qt::StringLiterals;
 
 // Settings keys
-constexpr auto DirPlaylist                  = "␟DirBrowserPlaylist␟";
 constexpr auto DirBrowserIconsKey           = u"DirectoryBrowser/Icons";
 constexpr auto DirBrowserDoubleClickKey     = u"DirectoryBrowser/DoubleClickBehaviour";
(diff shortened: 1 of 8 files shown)


ludouzi:
However, because the original settings keys are now treated as the default widget settings, this means existing users would have no means to restore the 'factory' defaults.

Easy way around this is to expose a `factoryDefaults` method and add a button to the config dialog to restore them.

ludouzi:
Widget config dialogs now expose 3 actions from the `Restore defaults` button:

1. Restore saved defaults - Restore previously saved defaults to this widget instance
2. Restore factory defaults - Restore the factory defaults for this widget instance
3. Clear saved defaults - Clear any saved defaults for all instances of this widget

PR #893, opened 2026-03-12
Fix: Show directory controls and location hiding after being enabled

Birbwell:
The conditionals for whether or not to display the settings were:
```
if(enabled && !m_upDir && ..) {
    // show controls
} else {
    // hide controls, even if they should be enabled (once m_upDir et al. is defined)
}
```
which would enable it right when you enable the setting, but immediately disable it due to m_upDir being defined after initially enabling it.

Fixed it by making it so the else block only runs if `enabled` is false. it's now 
```
if(enabled && !m_upDir && ..) {
    // show controls
} else if (!enabled) {
    // hide controls, only if disabled
 }
```

### DIFF
--- a/src/gui/dirbrowser/dirbrowser.cpp
+++ b/src/gui/dirbrowser/dirbrowser.cpp
@@ -58,9 +58,6 @@
 #include <QTreeView>
 #include <QUndoCommand>
 
-// TEMP
-#include <iostream>
-
 using namespace Qt::StringLiterals;
 
 // Settings keys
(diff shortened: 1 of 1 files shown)


Birbwell:
I realize it'd probably be better to provide a formal explanation of the issue in addition to the fix.

Basically, with the new UI changes for the dirbrowser settings, if you set to show the directory controls and location, you can briefly see it after hitting apply when enabling them, but they re-hide the moment the window closes.

This is because the setting to show them checks `If (setting enabled AND buttons _do not_ exist): show buttons. else: hide buttons.` I just set it so that the `else` clause also checks if the setting has been disabled before hiding the buttons, since otherwise it will hide them no matter if the setting is enabled/disabled as long as the buttons currently exist.

Issue #887, opened 2026-03-08
gnome-shell (very) rarely freezes when stopping playback


lstrsrt:
**Describe the bug**
This one is kind of tricky. I've been experiencing freezes in GNOME when shell uptime exceeds a couple of days and I pause/stop playback in fooyin. In other words I press pause and the screen stops updating with all apps becoming unresponsive (the only thing that still works is the cursor). This has been happening for at least two months. Switching to a different tty does not work anymore in this state, you have to Alt+Ctrl+Shift+Esc to get back to the login manager and then restart gdm (which restarts gnome-shell) to fix the freeze. Before this, gnome-shell is maxing out one core at 100% in htop.

**To Reproduce**
Leave GNOME running for >=2 days and pause/stop fooyin. The bug is very infrequent, sometimes it takes much longer to trigger. I am using two GNOME extensions ("Blur my Shell" and "AppIndicator and KStatusNotifierItem Support") right now, but I'll disable them for the time being to see if I can reproduce the bug without them.

**System:**
 - OS: Arch Linux
 - DE: GNOME 49.4
 - fooyin version: `f9b0f92`
 - Audio server: Pipewire

**Additional context**
I was able to reproduce this on the old and also the newly rewritten engine. The difference is that with the old engine, playback stopped before the freeze. Now it still plays, just stuttering and freezing a lot. I think there must be some kind of weird interaction bug between fooyin and GNOME as I have not had this happen on any other desktop environment or with any other app yet.

ludouzi:
That sounds very weird. I'll install GNOME to see if I can reproduce this too.

This reminds me of a few times in the past when my Hyprland session slowed to a crawl and then froze, which required a forced reboot. I had a few VMs running so I assumed it was those, but I also remember fooyin was open and playing too.

Could just be a coincidence, but maybe not after seeing your report. Definitely worth looking into further now.

ludouzi:
Do you have the MPRIS plugin enabled? 

Since you've seen it with both engines, and it's specific to GNOME, it seems more likely to be caused by system interaction over DBUS than anything playback related. I'll have a look through the MPRIS implementation.

lstrsrt:
Yeah the MPRIS plugin is enabled. DBUS interaction sounds like another possible bug source, I hadn't thought about that.

Lakis01:
[...]

PR #896, opened 2026-03-14
[mpris] Fix two MPRIS issues

lstrsrt:
Been reading the MPRIS plugin code to track down #887. I haven't been able to find the real problem so far but here are two other fixes. Note, I was calling `unregisterObject()` on shutdown but it doesn't seem to affect anything so I left it out.


ludouzi:
Thanks. I've been trying to reproduce #887 on my end but no luck yet either.

PR #908, opened 2026-03-20
[gui] ElidedLabel: Use QFontMetrics::size() instead of boundingRect()

lstrsrt:
`boundingRect()` does not account for spaces, so having a bunch of them in the status bar will make the size hint too conservative and cause text elision to happen early. Also reproducible with certain font and text combinations, for example here is Droid Sans with %playtime%:

image


ludouzi:
Good catch! I don't usually show track selection in the status bar, so it must have slipped past...

PR #909, opened 2026-03-20
[gui] ExpandedTreeView: Fix artifacts in horizontal scrolling

lstrsrt:
Update the viewport instead of just scrolling it.

Fixes this fun rendering bug:

[Screencast From 2026-03-20 18-45-47.webm](https://github.com/user-attachments/assets/719b32ce-ee8c-4e84-9804-c8a9ea3966d9)

The header text now stays at a constant position; we could move it along with the entries but keeping it in view looks better to me.


ludouzi:
Agreed; that looks better to me as well.

PR #910, opened 2026-03-20
[gui] Selection Info: Improve view resizing

lstrsrt:
`InfoView::resizeView()` always uses `header()` to change the layout. When "Show header" is disabled, the column length won't change because the header is hidden and `adjustSize()` won't do anything (this is relevant when toggling the horizontal scrollbar). My fix is to calculate the column length and resize manually. Also use viewport width instead of geometry width so there isn't a small leftover margin that keeps the view horizontally scrollable.


ludouzi (approved):
Looks good!

PR #914, opened 2026-03-21
[gui] SeekContainer: Resize labels dynamically

### LINKED ISSUE #493: Seekbar labels are cut off on tracks longer than 60min
**Describe the bug**
The seekbar labels get cut off if the duration or remaining duration of the track is longer than an hour, see screenshot.
It seems there is not enough margin to accommodate hh:mm:ss instead of mm:ss. 

**To Reproduce**
Enable labels of the seekbar and play a track that is longer than 60mins.

**Screenshots**

Image

**System:**
 - OS: Arch Linux
 - fooyin version: Version: 0.8.1
Based on Qt 6.8.2 (GCC 14.2.1 20250207, x86_64)

lstrsrt:
Let's close #493 before it turns one year old.

Delete `updateLabelSize` and handle sizes in `updateLabels` by checking if the string length has changed.
Also update the labels in `setElapsedTotal` so toggling "Show elapsed total" takes effect immediately when we're paused/stopped. Saves 10 lines of code!

### DIFF
--- a/include/gui/widgets/seekcontainer.h
+++ b/include/gui/widgets/seekcontainer.h
@@ -47,9 +47,6 @@ signals:
     void elapsedClicked();
     void totalClicked();
 
-protected:
-    void showEvent(QShowEvent* event) override;
-
 private:
     std::unique_ptr<SeekContainerPrivate> p;
 };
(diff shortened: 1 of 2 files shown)


ludouzi:
Nice!

PR #922, opened 2026-03-23
[gui] StatusWidget: Update playing text on pause

lstrsrt:
Somewhat counterintuitively, we also want to update the playing text on pause because the user may use a playback state dependent variable in the status script -- for example `%ispaused%`, which is how I noticed this.


ludouzi (approved):

ludouzi:
Thanks!

PR #924, opened 2026-03-25
[rgscanner] FFmpeg backend: Fix double free and clean up code

lstrsrt:
Running fooyin in Valgrind, I was getting errors originating from `~FFmpegContext()` after running an RG scan with the FFmpeg backend. The reason is that `avfilter_graph_free()` [already takes care](https://github.com/FFmpeg/FFmpeg/blob/master/libavfilter/avfiltergraph.c#L128) of freeing the `AVFilterContext`s associated with it, so we end up double freeing them after `ReplayGainFilter::filterGraph` gets destroyed.

[Similarly](https://github.com/FFmpeg/FFmpeg/blob/master/libavfilter/graphparser.c#L220-L223), we don't need to call `avfilter_inout_free()`. So remove the unused smart pointer deleters and add myself to the copyright header since I originally authored this file.

### DIFF
--- a/src/plugins/rgscanner/ffmpegscanner.cpp
+++ b/src/plugins/rgscanner/ffmpegscanner.cpp
@@ -166,7 +166,7 @@ ReplayGainFilter initialiseRGFilter(const Fooyin::AudioFormat& format, bool isPl
 
     AVFilterInOut* inputs   = nullptr;
     const auto filterParams = QString{u"ebur128=peak=%1,anullsink"_s}.arg(truePeak ? "true"_L1 : "sample"_L1);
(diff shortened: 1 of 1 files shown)


ludouzi:
Also, on the second point: we use `avfilter_graph_parse_ptr` rather than `avfilter_graph_parse`, in which the docs explicitly state we need to delete using `avfilter_inout_free`:

https://github.com/FFmpeg/FFmpeg/blob/238df21a4f94e30fb3225246db025e410e897965/libavfilter/avfilter.h#L795-L814

lstrsrt:
I think we can just switch over to `avfilter_graph_parse` then? We don't use `inputs` or `outputs` for anything else. I tried changing the call and calculation still worked fine.

ludouzi:
Yeah, switching over to `avfilter_graph_parse` should be fine, since we only ever use one output node.

PR #927, opened 2026-03-25
[gui] Add external drag support for playlist tracks

donutinit:
# I don't know
When I used windows I relied on foobar2000 to manage my sfx collection.
Now on linux the best fucking alternative is fooyin and it doesn't have the function to drag the file link into another program. I already did a PR #788 but didn't like the quality of code and since it was my first pr I deleted it lmao.

## Summary
- expose dragged playlist tracks as external file URLs and plain-text paths
- keep internal fooyin drag and drop working by prioritizing fooyin MIME data before generic URLs
- add tests for external track MIME data generation

## What changed
This change makes playlist track drag-and-drop also expose real local file data for other apps.

Dragged tracks now include:
- `file://` URLs
- plain text local paths

So from user side it should behave much closer to dragging the same files from a file manager.

At the same time, internal fooyin drag and drop is still preserved. I had to make sure fooyin-specific MIME data is still preferred for internal drops, otherwise internal drag actions could be treated like external file drops just because URLs are now also present.

## Implementation notes
- add a helper to populate `QMimeData` with valid local file URLs and plain text paths
- skip invalid files, archive-backed tracks, and duplicates
- use that helper when building playlist drag MIME data
- keep internal playlist drops preferring fooyin MIME payloads first, and only use URLs as fallback
- add unit tests for valid files, duplicates, and invalid selections

## Validation
- configured with `cmake --preset debug -DBUILD_CCACHE=OFF`
- built with `cmake --build build-debug -j$(nproc)`
- passed `build-debug/run/bin/test_guiutils`
- `ctest --test-dir build-debug/tests --output-on-failure -j$(nproc)` still shows one unrelated existing failure:
  - `AudioLoaderTest.SavesAndRestoresStateAcrossInstances`

### DIFF
--- a/include/gui/guiutils.h
+++ b/include/gui/guiutils.h
@@ -28,12 +28,15 @@
 #include <QByteArray>
 #include <QModelIndexList>
 
+class QMimeData;
+
 namespace Fooyin {
 class MusicLibrary;
 class SettingsManager;
 
 namespace Gui {
 FYGUI_EXPORT TrackList tracksFromMimeData(MusicLibrary* library, QByteArray data);
+FYGUI_EXPORT void populateExternalTrackMimeData(const TrackList& tracks, QMimeData* mimeData);
 FYGUI_EXPORT TrackList sortTracksForLibraryViewerPlaylist(SettingsManager* settings, const TrackList& tracks);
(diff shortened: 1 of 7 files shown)


ludouzi:
Changes look good and works fine after testing. Thanks!

PR #932, opened 2026-03-27
Delete Functionality

Agecaltin:
FYI my first pull request.
I added a delete functionality. This allows tracks to be deleted (ie moved to Trash) from the playlist view.

What works:
Right click on one or multiple selected tracks, under File operations select Delete. A confirmation window pops up asking if you want to delete the track(s). It has a checkbox to not ask again. Just hitting enter cancels the delete. This can be reset in the Settings under Playlist at the bottom. Default is that the confirmation is asked.
The Delete functionality can be assigned to a keyboard shortcut too.

I tested it a bit on my own but please have a thorough look at my code.

There are some Issues open requesting this feature I believe.

Thanks, Sebastian

### DIFF
--- a/src/plugins/fileops/fileopsplugin.cpp
+++ b/src/plugins/fileops/fileopsplugin.cpp
@@ -19,6 +19,7 @@
 
 #include "fileopsplugin.h"
 
+#include "fileopsconfigdialog.h"
 #include "fileopsdefs.h"
 #include "fileopsdeletedialog.h"
 #include "fileopsdialog.h"
@@ -40,6 +41,32 @@
 #include <QThread>
 
 using namespace Qt::StringLiterals;
+
+namespace {
+bool canOperateOnTracks(const Fooyin::TrackList& tracks)
+{
+    return !tracks.empty()
+        && std::ranges::all_of(tracks, [](const Fooyin::Track& track) { return !track.isInArchive(); });
+}
+
+class FileOpsPluginSettingsProvider : public Fooyin::PluginSettingsProvider
+{
+public:
+    explicit FileOpsPluginSettingsProvider(Fooyin::SettingsManager* settings)
+        : m_settings{settings}
+    { }
+
+    void showSettings(QWidget* parent) override
+    {
+        auto* dialog = new Fooyin::FileOps::FileOpsConfigDialog(m_settings, parent);
+        dialog->setAttribute(Qt::WA_DeleteOnClose);
+        dialog->show();
+    }
+
+private:
+    Fooyin::SettingsManager* m_settings;
+};
+} // namespace
 
 namespace Fooyin::FileOps {
 FileOpsPlugin::FileOpsPlugin()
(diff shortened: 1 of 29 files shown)


Agecaltin:
I hope I get notified via email for any replies because i am rarely here

ludouzi:
Thanks. I don't have time to go through this before the next release today, but another release will likely follow soon after so it can be added for then.

Agecaltin:
@ludouzi Oh, awesome! I am trying to fix the Clang-format issue. And no time pressure.

hubbaredneck:
Thank you @Agecaltin , I have been wanting this feature for some time now, one of the few things remaining I still miss from foobar2k.

To confirm this is only for the right-click context menu in the playlist view?
It will not be possible to assign a hotkey to delete the currently playing song and skip to the next song in the playlist with this?

src/plugins/fileops/fileopsworker.cpp
```diff
@@ -167,11 +182,17 @@ void FileOpsWorker::run()
         m_library->updateTrackMetadata(m_tracksToUpdate);
     }
 
+    if(!m_tracksToDelete.empty()) {
+        m_library->removeUnavailbleTracks();
+    }
```
ludouzi: This triggers a full library rescan, which may be slow on network drives, and if the user has unrelated offline/network/archive-backed files at the time they delete one local track, those unrelated entries may get removed too.

src/plugins/fileops/fileopsplugin.cpp
```diff
@@ -122,5 +129,47 @@ void FileOpsPlugin::recreateMenu()
...
+            auto* worker = new FileOpsWorker(m_library, tracks, m_settings);
+            auto* thread = new QThread(this);
+            worker->moveToThread(thread);
+
+            QObject::connect(worker, &FileOpsWorker::simulated, worker,
+                             [worker]() { QMetaObject::invokeMethod(worker, &FileOpsWorker::run); });
+            QObject::connect(worker, &Worker::finished, thread, &QThread::quit);
+            QObject::connect(worker, &Worker::finished, this, [this]() {
+                if(auto* removeCmd = m_actionManager->command(Constants::Actions::Remove)) {
+                    removeCmd->action()->trigger();
```
ludouzi: `FileOpsWorker::finished` is emitted regardless of whether or not the file(s) were successfully deleted from disk. If the file wasn't actually deleted, this will still remove it from the playlist, which is not ideal.

In addition, the File Ops menu is attached to the shared track-selection context, and that same context menu is used from the playlist, library tree, and filters. After delete, the remove action is triggered; but when the current context has no `Edit.Remove` override, the command proxy keeps its previous action instead of clearing it. Deleting from a library viewer can then trigger a stale playlist/queue/organiser remove action.

ludouzi:
Thanks for the PR!

Most of the changes look good, though there are some issues which need addressing. I'd like to get this merged in time for a new release by tomorrow, so I'll make some changes, especially since the `MusicLibrary` interface will need to be added to.

ludouzi:
I've also added a FileOps plugin config dialog to host the `Confirm before deleting tracks` setting, keeping it self-contained within the plugin.

ludouzi (approved):

PR #945, opened 2026-03-29
[core,wavebar] Implement support for multichapter-files

thatguy7:
Full disclosure, I have not written C++ in forever. I generated these changes with AI - so sorry for the noise.

I created two example "books" for testing: [TheLibrary.zip](https://github.com/user-attachments/files/26326890/TheLibrary.zip)

- **TheLibrary.m4b**: AAC VBR stream in a mp4 container (so literally the same thing as a .m4a or .mp4 file, just different file extension). .m4b is pretty much the default format for long-form chaptered audio. E.g. used by Audible. AAC is a well established format, so most players and software works with it just fine.
- **TheLibrary.mka**: Opus stream in a Matroska container. .mka is more of a niche. Not sure many people use it, but they probably should. For stuff like podcasts and audiobooks Opus makes sense as it sounds mostly transparent for speech at very low bitrates. .mka as a container is more flexible than .mp4. But also less supported on hardware - though I would argue e.g. FFmpeg has better support for .mka/.mkv than .mp4.

In theory, FFmpeg can generate other containers as well. E.g. a .mp3 with chapters, but these never worked well for me so I never used them - though could generate them for testing as well.

The .m4b seems to work well with these changes, but I had to set:
Playback/Library/Tag readers - moved FFmpeg to the top.
Playback/Library/FFmpeg/Enable all supported extension - enabled

The .mka file plays as well, but there are issues with the WaveBar widget. After the first chapter the waveform is no longer in sync with the audio.

This relates to this issue: https://github.com/fooyin/fooyin/issues/694
@kode54  made a draft for the same issue here: https://github.com/fooyin/fooyin/pull/790

Files downloaded with yt-dlp, as suggested here: https://github.com/fooyin/fooyin/pull/790#issuecomment-3657954715  by @tmtom  work just fine for me.

I hope this can contribute to a proper implementation of the feature.

### DIFF
--- a/src/core/engine/input/ffmpeg/ffmpeginput.cpp
+++ b/src/core/engine/input/ffmpeg/ffmpeginput.cpp
@@ -946,7 +946,8 @@ public:
             m_ioContext.reset();
         }
 
-        m_stream = {};
+        m_stream       = {};
+        m_chapterCount = 0;
     }
 
     bool setup(const AudioSource& source)
@@ -962,12 +963,21 @@ public:
         }
 
         m_stream = Utils::findAudioStream(m_context.get());
-        return m_stream.isValid();
+        if(!m_stream.isValid()) {
+            return false;
+        }
+
+        if(m_context->nb_chapters > 1) {
+            m_chapterCount = static_cast<int>(m_context->nb_chapters);
+        }
+
+        return true;
     }
 
     IOContextPtr m_ioContext;
     FormatContextPtr m_context;
     Stream m_stream;
+    int m_chapterCount{0};
 };
 
 FFmpegReader::FFmpegReader()
@@ -990,6 +1000,11 @@ bool FFmpegReader::canReadCover() const
 bool FFmpegReader::canWriteMetaData() const
 {
     return false;
+}
+
+int FFmpegReader::subsongCount() const
+{
+    return p->m_chapterCount > 1 ? p->m_chapterCount : 1;
 }
 
 bool FFmpegReader::init(const AudioSource& source)
@@ -1062,6 +1077,51 @@ bool FFmpegReader::readTrack(const AudioSource& /*source*/, Track& track)
     AVDictionaryEntry* tag{nullptr};
     while((tag = av_dict_get(p->m_context->metadata, "", tag, AV_DICT_IGNORE_SUFFIX))) {
         parseTag(track, tag);
+    }
+
+    const int subsong = track.subsong();
+    if(p->m_chapterCount > 1 && subsong >= 0 && subsong < p->m_chapterCount) {
+        const AVChapter* chapter = p->m_context->chapters[subsong];
+
+        const uint64_t startMs = av_rescale_q(chapter->start, chapter->time_base, TimeBaseMs);
+        const uint64_t endMs   = av_rescale_q(chapter->end, chapter->time_base, TimeBaseMs);
+
+        // Container title becomes album (e.g. audiobook title)
+        if(!track.title().isEmpty()) {
+            track.setAlbum(track.title());
+        }
+
+        track.setOffset(startMs);
+        track.setDuration(endMs - startMs);
+
+        // Read chapter-level metadata
+        AVDictionaryEntry* chapterTag{nullptr};
+        QString chapterTitle;
+        QString chapterTrackNumber;
+
+        while((chapterTag = av_dict_get(chapter->metadata, "", chapterTag, AV_DICT_IGNORE_SUFFIX))) {
+            if(strcasecmp(chapterTag->key, "title") == 0) {
+                chapterTitle = convertString(chapterTag->value);
+            }
+            else if(strcasecmp(chapterTag->key, "track") == 0 || strcasecmp(chapterTag->key, "part_number") == 0) {
+                chapterTrackNumber = convertString(chapterTag->value);
+            }
+        }
+
+        if(!chapterTitle.isEmpty()) {
+            track.setTitle(chapterTitle);
+        }
+
+        if(!chapterTrackNumber.isEmpty()) {
+            track.setTrackNumber(chapterTrackNumber);
+        }
+        else {
+            track.setTrackNumber(QString::number(subsong + 1));
+        }
+
+        track.setTrackTotal(QString::number(p->m_chapterCount));
+
+        track.setExtraProperty(u"CHAPTER"_s, u"1"_s);
     }
 
     return true;
--- a/src/core/engine/input/ffmpeg/ffmpeginput.cpp
+++ b/src/core/engine/input/ffmpeg/ffmpeginput.cpp
@@ -1098,6 +1098,8 @@ bool FFmpegReader::readTrack(const AudioSource& /*source*/, Track& track)
         AVDictionaryEntry* chapterTag{nullptr};
         QString chapterTitle;
         QString chapterTrackNumber;
+        float chapterRGTrackGain{Fooyin::Constants::InvalidGain};
+        float chapterRGTrackPeak{Fooyin::Constants::InvalidPeak};
 
         while((chapterTag = av_dict_get(chapter->metadata, "", chapterTag, AV_DICT_IGNORE_SUFFIX))) {
             if(strcasecmp(chapterTag->key, "title") == 0) {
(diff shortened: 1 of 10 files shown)


kode54:
Broken: Every track is the first subsong.

Edit: Disregard, I had to delete my fooyin.db to fix the cached track info.

Meanwhile, could you look into fixing album titles? FFmpeg supplies the "album" tag named "title" for chapterized files.

thatguy7:
I am interested in getting this going, but I caught a cold and will be non-responsive over the next days.

kode54:
> I am interested in getting this going, but I caught a cold and will be non-responsive over the next days.

I apologize for my abrupt feedback. I wish you a speedy recovery, if only because I know how annoying being sick is. On the other hand, I hope you take it easy.

I confirmed the chapter support you coded works fine, as long as I wipe my database of any entries generated by tracks I already probed against my existing implementation. Otherwise, every track is missing the correct subsong number, so they all get the same Offset and Length as the first track, and effectively play the first track.

Your implementation seems better than mine, at least. Though it does need propagating the input "title" field to the "album" tag before overriding it with the track "title" value, since FFmpeg multi-chapter files store the album title in the global "title" field.

It would also be neat to have support for the ReplayGain tags from FFmpeg. FFmpeg implements those as `REPLAYGAIN_GAIN` and `REPLAYGAIN_PEAK` fields, either on the global metadata dictionary, or the per-chapter dictionaries. Possibly also on the per-program dictionaries, but I don't know how those work for audio files.

kode54:
Y'all okay there? I hope you recovered by now. :D

thatguy7:
Doing way better now, but it got me real good this time, still coughing. I intend to get back to this over the weekend.

Just to clarify what tags are supposed to be overriden/propagated I created this diagram, that is roughly my understanding of how most containers work:
Untitled Diagram

We have tags at the global/container level, at the stream level and at the chapter level.

Right now, the tags at container/stream level are read in and used as the basis for all tracks/subsongs. Then the chapters-level tags are read and override the container level tags on a per chapter/subsong basis.

What you suggest, is to take the container level **title** tag and use it to override the container level **album** tag, and only after replace the container level **title** tag with the chapters level **title** tag - correct? Do we override the **album** tag only if it does not exist or always?

I just verified that ffmpeg lets me put e.g. **REPLAYGAIN_TRACK_GAIN** as a chapter level tag, so I guess it is possible to put there anything. So analogously to propagating **title** from chapter for each subsong, we can also propagate the replay gain tags. Actually, it kind of would make sense to propagate all tags defined at chapter level to the subsong.

thatguy7:
> I confirmed the chapter support you coded works fine, as long as I wipe my database of any entries generated by tracks I already probed against my existing implementation. Otherwise, every track is missing the correct subsong number, so they all get the same Offset and Length as the first track, and effectively play the first track.

Not sure I understand this correctly. In the UI it seems to me like the subsong number is set correctly for all subsongs but the first one - which does not seem to get one. When I skip from chapter to chapter it seeks to the correct position and does not play the first track for all subsongs.

kode54:
> > I confirmed the chapter support you coded works fine, as long as I wipe my database of any entries generated by tracks I already probed against my existing implementation. Otherwise, every track is missing the correct subsong number, so they all get the same Offset and Length as the first track, and effectively play the first track.
> 
> Not sure I understand this correctly. In the UI it seems to me like the subsong number is set correctly for all subsongs but the first one - which does not seem to get one. When I skip from chapter to chapter it seeks to the correct position and does not play the first track for all subsongs.

This was just a problem when I was migrating an install from my implementation to yours.

thatguy7:
I have rebased my changes onto current master and cleaned up the commit history a bit:

- I removed the explicit formatting commits and ammended the formatting to the preceding commits - otherwise it made rebasing on current master annoying.
- I removed the commit that prevented the title -> album tag propagation.

Otherwise the changes in the commits remained untouched.

Not sure if it was there to begin with or added to master after the last time I touched the code, but there is a handy `parseTag` function that makes reading all tags supported at container level also easily done at chapter level. So now all tags supported at container level, which as I understand it is literally anything, is now supported at chapter level.

src/core/engine/input/ffmpeg/ffmpeginput.cpp
```diff
@@ -1052,6 +1067,36 @@ bool FFmpegReader::readTrack(const AudioSource& /*source*/, Track& track)
...
+        AVDictionaryEntry* chapterTag{nullptr};
+        while((chapterTag = av_dict_get(chapter->metadata, "", chapterTag, AV_DICT_IGNORE_SUFFIX))) {
+            parseTag(track, chapterTag);
+        }
+
+        // Ensure chapter timing and totals are authoritative
+        track.setDuration(endMs - startMs);
+        track.setTrackTotal(QString::number(p->m_chapterCount));
+
+        track.setExtraProperty(u"CHAPTER"_s, u"1"_s);
```
thatguy7, line 1097: I really dislike this. In order to mark the track as "this is a chapter" this property is added. But it is also shown in the UI, which it should not be.

The track should get a track.isChapter(), but I am not sure of how to add it cleanly. I assume one would need to check it most of the times when track.hasCue() is checked. So maybe it should be more tightly integrated with that?

src/core/engine/input/ffmpeg/ffmpeginput.cpp
```diff
@@ -917,12 +918,21 @@ class FFmpegReaderPrivate
         }
 
         m_stream = Utils::findAudioStream(m_context.get());
-        return m_stream.isValid();
+        if(!m_stream.isValid()) {
+            return false;
+        }
+
+        if(m_context->nb_chapters > 1) {
```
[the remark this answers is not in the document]
thatguy7: I changed it to` >= 1`.

kode54:
As I said before, FFmpeg doesn't really use "REPLAYGAIN_TRACK_GAIN" or "REPLAYGAIN_ALBUM_GAIN".

It explicitly uses "REPLAYGAIN_GAIN" and "REPLAYGAIN_PEAK", without the "_ALBUM" or "_TRACK". The location of the tags is how you determine whether they're album or track tags.

src/core/engine/input/ffmpeg/ffmpeginput.cpp
ludouzi: Since c4db0202cc1bb0745dc63b7aa7c98e3c4b26388f, track properties prefixed with `_` are hidden from the UI, so something like `_CHAPTER` would work fine here.

Adding `Track::isChapter` would require adding an additional DB column to persist it.

ludouzi:
I've made some changes after considering all of the above.

- Chapter state is stored as an internal `_CHAPTER` property
- Chapter 1 now keeps it's own unique identify (previously it could share it with the whole file as `offset == 0`)
- Title is read as the album title in `parseTag` for multi-chapter files in the global scope
- `REPLAYGAIN_GAIN/PEAK` are mapped by metadata scope: global -> album, chapter/stream -> track

I also fixed the reload path so reloading chapter tracks preserves the existing track identity.

ludouzi:
I haven't encountered any issues while testing this PR with a few `m4b` and `mka` files, and I don't see any blockers, so I will likely merge this today.

The only issue I have is reader priority. `FFmpeg` currently needs a higher priority than `TagLib` for multi-chapter files in formats that both readers support, which means the TagLib reader never gets used. The simplest fix would be a setting that prefers FFmpeg for known chapter-capable formats. A more robust approach would be to query reader candidates for those files and prefer the reader that reports the highest `subsongCount`.

On a related note, `TagLib` has added Matroska support in recent releases, along with initial chapter support. I'll look at adding that to the TagLib reader at some point as well.

thatguy7:
It works for my use cases and my files well but not perfectly. There is still a bug I can replicate with any file I try (including the ones I attached):

When the track switches automatically (e.g. Track 1 finishes so Track 2 plays after), I can not seek in Track 2. Not using the progress bar, and not in the wavebar widget.

If I pick Track 2 from the playlist, I can seek in the track with no issues. So the issue is only there, when the track starts playing automatically as part of the playlist.

So easy to replicate:

- load chaptered file
- playback any chapter (I double-click on a track in the UI)
- seek to the last seconds of chapter (or just wait till the chapter finishes)
- once playback jumps to the next chapter in the playlist, seeking no longer works

thatguy7:
No, scratch that.

The seeking works, but the UI breaks.

The Seekbar and wavebar widget do not show the position correctly. So when I seek to the middle of track 2, that is what fooyin plays. But the progressbar and wavebar jump to position 0. And once the seekbar reaches the end, the UI no longer jumps to the next track.

ludouzi:
Looks like I missed that!

After seeking, the active stream position is re-synced relative to the current chapter, but the stream origin was not being updated. Fixed in https://github.com/fooyin/fooyin/pull/945/commits/e4fd896f47d38ee01f29367482c1cb403a6c60e1.

thatguy7:
Amazing. Already fixed.

No more complaints from my side. It works.

Though I must admit I have not tested anything Replay-Gain related.

ludouzi:
I’ve confirmed that `REPLAYGAIN_*_GAIN` and `REPLAYGAIN_*_PEAK` work.

I haven’t tested `REPLAYGAIN_GAIN`/`REPLAYGAIN_PEAK` directly, but they go through the same path and the logic is correct, so they should behave the same.

PR #975, opened 2026-04-03
Some small enhancements for v0.10.3

asflower:
- Add default support for matching comment fields in scripting
- Add lyrics negative offset support
- Fix an issue where icons were not displayed correctly in X11 environments


ludouzi:
Thanks! It looks like Wayland doesn't use/care about `StartupWMClass`, so that change should be fine for both X11 and Wayland.

PR #1019, opened 2026-04-10
Fix wavebar option labeling

### LINKED ISSUE #1017: Misnamed wavebar config option?
In the following lines, there is a config value for the wavebar that is called "Show elapsed total":
https://github.com/fooyin/fooyin/blob/6a63de9031e24dc47cf6875c7a3ef43294d8123a/src/plugins/wavebar/wavebarconfigwidget.cpp#L44
https://github.com/fooyin/fooyin/blob/6a63de9031e24dc47cf6875c7a3ef43294d8123a/src/plugins/wavebar/wavebarwidget.cpp#L517
https://github.com/fooyin/fooyin/blob/6a63de9031e24dc47cf6875c7a3ef43294d8123a/src/gui/controls/seekbar.cpp#L357

When enabled, the wavebar shows the elapsed time (left) and the **remaining time** (right). See the following picture, where the setting is enabled (pay attention to the right of the wavebar):

Image

In contrast, while disabled, it shows the elapsed time (left) and the **track duration** (right):

Image

Therefore the setting should be named "Show remainder" or something like that?

numerfolt:
Fixes https://github.com/fooyin/fooyin/issues/1017#issuecomment-4219355607 by renaming all occurances of the wavebar option `Show elapsed total` to `Show remaining time`.

### DIFF
--- a/src/gui/controls/seekbar.cpp
+++ b/src/gui/controls/seekbar.cpp
@@ -354,7 +354,7 @@ void SeekBar::contextMenuEvent(QContextMenuEvent* event)
                      [this](bool checked) { m_container->setLabelsEnabled(checked); });
     menu->addAction(showLabels);
 
-    auto* showElapsed = new QAction(tr("Show elapsed total"), menu);
+    auto* showElapsed = new QAction(tr("Show remaining time"), menu);
     showElapsed->setCheckable(true);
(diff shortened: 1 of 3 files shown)


ludouzi (approved):
Thanks!

Issue #496, opened 2025-03-24
[Feature Request] Context -> File -> Copy Path


JkktBkkt:
We already have "Open containing folder" but for quite a few operations operating on the path directly makes more sense, which in current state means you either need to copy from file properties or open the folder and copy path from the file manager UI. 
I'd love to see 2 entries: "copy file path" and "copy directory path"

PR #1020, opened 2026-04-10
[gui] Add "Copy file location" action

lstrsrt:
Addresses #496.

I don't know if we want another action to copy just the directory, the context menu looks pretty long to me as it is. Maybe under a submenu?

### DIFF
--- a/src/gui/trackselectioncontroller.cpp
+++ b/src/gui/trackselectioncontroller.cpp
@@ -635,7 +635,7 @@ void TrackSelectionControllerPrivate::copyLocation(const TrackSelection& selecti
     for(const auto& track : selection.tracks) {
         paths.append(track.prettyFilepath());
     }
-    QApplication::clipboard()->setText(paths.join(QLatin1String{Constants::NewLine}));
+    QApplication::clipboard()->setText(paths.join("\n"_L1));
 }
 
 void TrackSelectionControllerPrivate::searchArtwork(const TrackSelection& selection, bool quick) const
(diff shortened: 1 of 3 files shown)


src/gui/trackselectioncontroller.cpp
```diff
@@ -610,6 +625,19 @@ void TrackSelectionControllerPrivate::openFolder(const TrackSelection& selection
...
+{
+    if(selection.tracks.empty()) {
+        return;
+    }
+
+    QStringList paths;
+    for(const auto& track : selection.tracks) {
+        paths.append(track.prettyFilepath());
+    }
+    QApplication::clipboard()->setText(paths.join(QLatin1String{Constants::NewLine}));
```
ludouzi: The `NewLine` constant isn't necessary here (or in most calls involving QString), as Qt normalises `\n` depending on the platform.

In this case, `QWindowsMimeText::convertFromMime` handles it.

ludouzi (changes requested):

src/gui/trackselectioncontroller.cpp
lstrsrt: Fixed. I removed `Constants::NewLine` as well because it's not used anywhere else.

ludouzi:
Another action to copy the directory could be useful, but that would definitely add too much clutter to the context menu and I don't suspect many people would use it.

I've got plans to make the track selection menu customisable in terms of what actions and submenus are included, so it could be considered then.

ludouzi (approved):
Thanks!

PR #1030, opened 2026-04-13
Extend Delete file functionality

l-raider:
This extends the delete file functionality added in https://github.com/fooyin/fooyin/pull/932 with the following:
- add keyboard mnemonics to delete confirmation dialog
- show filepath in single-track delete confirmation dialog (I found just showing the track Title was a little lacking in information)
- add Delete Currently Playing File shortcut

Delete Currently Playing File does not trigger if file is an archive and only if there is an active track playing or paused.
Shortcut/Hotkey is unassigned by default.

This is to mimic the following hotkey assignment in foobar2000: 
Type:  [ Context / Now Playing ], Action: Delete

I used LLM for assistance as I am not very familiar with QT. (This will be my first QT related PR, let me know if anything is missing...)


l-raider:
Aborting this for now, moving this functionality into a plugin instead and will create a PR later when the code is a little more battle tested: https://github.com/l-raider/fooyin-plugins/tree/main/shortcutextender

PR #1049, opened 2026-04-21
[utils] Calculate rich text size with QFontMetrics::size()

lstrsrt:
The bug fixed in #908 reappeared when `ElidedLabel::richTextNaturalSize()` was switched over to use `measureRichText()`. Squash it again, this time in `measureRichTextLine()`.

### DIFF
--- a/src/gui/scripting/richtextutils.cpp
+++ b/src/gui/scripting/richtextutils.cpp
@@ -33,9 +33,10 @@ RichTextMetrics measureRichTextLine(const RichText& richText, const QFont& baseF
     for(const auto& block : richText.blocks) {
         const QFont font = resolvedRichTextFont(block.format, baseFont);
         const QFontMetrics fm{font};
+        const auto size = fm.size(Qt::TextSingleLine, block.text);
 
-        metrics.width += fm.horizontalAdvance(block.text);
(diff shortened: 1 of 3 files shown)


ludouzi:
Oops!

The reason this is an issue for `ElidedLabel` is that it uses `QFontMetrics::size()` in its paint path, while the shared `measureRichText()` helper and the other callers use `horizontalAdvance()` consistently for layout/sizing.

I don’t think the right fix is to change `measureRichText()` globally, though. `size()` is not a good general-purpose metric for rich-text layout when a line, or even a single word, is split into multiple styled blocks, as summing `size().width()` across those blocks will spread the text out:

image

As this looks specific to `StatusLabel` for the right-aligned selection label, I’m working on fixing it in `StatusWidget`, while also making sure that the right-hand text stays right-aligned:

image

ludouzi:
I just realised the right-aligned selection label doesn't even need text elision, so that's now been turned off. `ElidedLabel` also now uses `horizontalAdvance` so it's consistent with most other rich-text users.

`ElidedLabel` should still be looked at to report a more appropriate `sizeHint()` for the rich-text case so all elide modes can be used, but there is no longer any users of it which need both text elision and also align right, so that can be saved for later.

ludouzi (approved):

Issue #1048, opened 2026-04-21
(AAC) fooyin does not handle replaygain info correctly for files tagged using foobar2000 or rsgain


Sekai9:
**Describe the bug**
fooyin has issues handling replaygain metadata for AAC/M4A files tagged using foobar2000 or rsgain. Track and Album gain are missing so there is no gain correction, Peak data seems to be correctly parsed.

**To Reproduce**
Add and check the attached audio files to fooyin. They're already tagged with their respective scanners.

**Expected behaviour**
Proper gain correction

**Screenshots**

**System:**
 - OS: Fedora 43 KDE
 - fooyin version: 0.10.3

**Additional context**
A) fooyin correctly handles AAC files only when tagged by its own RG scanner. foobar2000 and Vanilla Music (Android) correctly handle RG data written by fooyin.

B) if you open the files with a hex editor, you can see that both foobar and rsgain write the gain as:
replaygain_track_gain....data........-10.40 dB  <--- tagged by foobar
REPLAYGAIN_TRACK_GAIN....data........-10.40 dB  <--- tagged by rsgain (the uppercase tags seems to cause another issue for fooyin... see "C")
while fooyin writes gain data as:
replaygain_track_gain....data........-10.4037
So that's probably the issue...

C) rsgain by default writes the tags in uppercase, this seems to cause fooyin to not display any info in ReplayGain section (Selection Info), the sample tagged by foobar shows the peaks correctly, but the gain is not correct (0.0dB), but it shows some info at least.
fooyin displays the RG tags as Extended Metadata for all samples.

[audio samples.zip](https://github.com/user-attachments/files/26915210/audio.samples.zip)

PR #1050, opened 2026-04-21
[core] TagLib: Handle MP4 gain strings with dB suffix

lstrsrt:
For #1048, the problem is that the MP4 tag handler overrides the RG values from the general tag reader, but it does not convert gain strings correctly when they have a dB suffix. Fix that by using `gainStringToFloat`.

Also write gain strings with a dB suffix to match foobar2000 and rsgain.

As a side note, do we want to always override the general tag reader here?

### DIFF
--- a/src/core/engine/input/taglibparser.cpp
+++ b/src/core/engine/input/taglibparser.cpp
@@ -563,6 +563,11 @@ float gainStringToFloat(const TagLib::String& gainString)
     return ok ? gain : Fooyin::Constants::InvalidGain;
 };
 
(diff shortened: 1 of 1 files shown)


lstrsrt:
@ludouzi Just noticed that you self-assigned the issue. Should I keep this PR open?

ludouzi:
Yeah, I started working on this last night. I'll close this, as I've just made a more comprehensive fix in https://github.com/fooyin/fooyin/commit/3e95658617832efb9ccbdb46d85f5612ed1ce968.

PR #1063, opened 2026-04-24
[gui] TabStackWidget: Use hovered tab index for name editor

lstrsrt:
When a tab in a tab stack is renamed through the context menu, the renaming will happen for the currently active tab instead of the tab that was actually hovered before the right click. Fixed by making `EditableTabBar::showEditor()` take an `index` argument and passing it down.


ludouzi (approved):
Thanks!

PR #1075, opened 2026-04-27
[vumeter] Don't set static layer DPR

lstrsrt:
When DE scaling is set to >100%, the legend overlay is drawn oversized:

[Screencast From 2026-04-27 23-35-13.webm](https://github.com/user-attachments/assets/f40f8030-7576-4f8f-b6ad-9e2120081c4e)

Fix by removing the call to `QPixmap::setDevicePixelRatio()`.


ludouzi (commented):
Good catch! Though removing `setDevicePixelRatio` is the wrong fix for this.

The static layer is still allocated as `targetSize * targetDpr`, so without setting the pixmap DPR, Qt treats it as a 1x pixmap. That fixes the rect source mismatch, but it makes the cached legend/grid render at 1x on high-DPI displays:

image

A better fix is to keep the DPR on the pixmap and scale the source rect:

image

I'll make some changes.

PR #1098, opened 2026-05-02
[core] TagLib: Make TDRC override TYER date tag

### LINKED ISSUE #787: Date tag handling
Just got fooyin and I'm enjoying it so far, it's pretty close to all I need from a music player. However, I've encountered one oddity: I have made a best effort to keep all my tags clean in ID3v2.4 format, but some of them have a duplicative TYER and TDRC field. Fooyin handles this by concatenating the fields, resulting in dates like "2005 2005", etc.

Image

Now, technically, with ID3v2.4, TYER was deprecated and replaced by TDRC, so these are technically improperly tagged; however, many tools still rely on the old TYER and I'm reticent to remove the tag as it would damage how my library looks in Jellyfin. Fooyin is the only media manager I've encountered this issue on, and I do feel that such a problem could probably be handled more gracefully. Instead of concatenating, my expectation would be that if both dates are the same, it simply shows one value. Non-matching dates could still concatenate or be handled more properly.

lstrsrt:
Closes #787.

If we spot both `TDRC` and `TYER` in the same file, let's use `TDRC` on ID3v2.4. FFmpeg does the opposite and uses `TYER` instead, but I think using the newer tag makes more sense. If we want both tag readers to have the same behavior to the user, we could of course switch it around.

Edit: Another possibility is to concatenate if the contents of `TDRC` and `TYER` don't match. But this might confuse users who don't know much about ID3 tags.

### DIFF
--- a/src/core/engine/input/taglibparser.cpp
+++ b/src/core/engine/input/taglibparser.cpp
@@ -1146,8 +1146,18 @@ void readId3Tags(const TagLib::ID3v2::Tag* id3Tags, Fooyin::Track& track)
     }
 
     const TagLib::ID3v2::FrameListMap& frames = id3Tags->frameListMap();
+    const unsigned int majorVersion           = id3Tags->header()->majorVersion();
 
-    if(id3Tags->header()->majorVersion() == 3) {
+    if(majorVersion == 4) {
+        // TDRC replaces TYER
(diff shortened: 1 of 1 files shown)


ludouzi (approved):

ludouzi:
Agreed; it's probably best to only use the value of the newer tag if both are present. Most users will only expect a single value for a field like `Date`.

PR #1101, opened 2026-05-03
[lyrics] Don't update lyrics widget if track changed during lyrics editing

### LINKED ISSUE #748: Lyrics Editor is not bound to the song you started editing when using Lyrics widget -> Edit lyrics
If you're editing a song's lyrics by right clicking in the Lyrics widget and selecting Edit lyrics, the Lyrics Editor that opens will not be bound to the song you started editing in, which means you may end up saving the lyrics of Song A to Song B's metadata.

The only way to access a song's lyrics through the Lyrics widget for a specific song is to start playing the song (unlike for example album artwork, which can be selected to be either the currently playing track or the highlighted track).
The playing song's lyrics can then be edited by right clicking the Lyrics widget and selecting Edit lyrics (a fairly recently implemented feature).

However, if you started editing and the highlighted track changes (the next song starts playing or you click on another song in your playlist), and you were to hit Apply at this point, the lyrics would be applied to the next song and not the one you were originally editing.

This does not happen when going to the Properties of a certain song and doing the editing in the Lyrics tab, because Fooyin is specifically referring to the lyrics of the track you opened the Properties for.

1. Select a song (Song A) and start editing the lyrics using the Lyrics widget - Edit lyrics
2. Select the next song (Song B) by either starting it from your playlist, skipping to the next track or letting the next track start playing at the end of Song A
3. The lyrics for Song A will stay open but when you hit Apply, will be written to Song B's metadata instead

It's to be expected that an opened Lyrics Editor for Song A would be tied to Song A's metadata instead of accessing the currently playing track Song B's metadata.

As stated before, this does not happen when going to Properties and editing lyrics in the specific song's Lyrics tab, because Fooyin is then targeting a specific song instead of opening a generic Lyrics Editor.

lstrsrt:
I believe #748 was already fixed at some point because I cannot reproduce it on my end.

The remaining problem is that if the current track changes while its lyrics are being edited, the lyrics view will still be changed for the new track, making it look like the lyrics were updated for the new track instead. This PR fixes that.

Closes #748.

By the way, is it intentional that the Lyrics tab in the properties dialog always uses the track's metadata lyrics?

### DIFF
--- a/src/plugins/lyrics/lyricswidget.cpp
+++ b/src/plugins/lyrics/lyricswidget.cpp
@@ -787,22 +787,30 @@ void LyricsWidget::handleSavedLyrics(const Track& track, const Lyrics& lyrics)
     changeLyrics(lyrics);
 }
 
-void LyricsWidget::changeLyrics(const Lyrics& lyrics)
+void LyricsWidget::changeLyrics(const Lyrics& lyrics, bool onEdit)
 {
-    m_currentLyrics = lyrics;
+    const bool fullUpdate = !onEdit || m_currentTrack.sameIdentityAs(*m_editingTrack);
+
+    if(fullUpdate) {
+        m_currentLyrics = lyrics;
+    }
 
     if(!lyrics.isLocal) {
         m_lyricsSaver->autoSaveLyrics(lyrics, m_currentTrack);
     }
 
     if(lyrics.isValid()) {
-        m_model->setLyrics(m_currentLyrics);
-        m_lyricsView->setDisplayString({});
+        if(fullUpdate) {
+            m_lyricsView->setDisplayString({});
+            m_model->setLyrics(m_currentLyrics);
+        }
         updateEdgeFadeState();
     }
     else {
-        m_lyricsView->setDisplayString(noLyricsDisplayText(m_currentTrack));
-        m_model->setLyrics({});
+        if(fullUpdate) {
+            m_lyricsView->setDisplayString(noLyricsDisplayText(m_currentTrack));
+            m_model->setLyrics({});
+        }
         m_lyricsView->setEdgeFadeEnabled(false);
     }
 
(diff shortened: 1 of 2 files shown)


src/plugins/lyrics/lyricswidget.cpp
```diff
@@ -795,7 +795,7 @@ void LyricsWidget::changeLyrics(const Lyrics& lyrics)
         m_lyricsSaver->autoSaveLyrics(lyrics, m_currentTrack);
     }
 
-    if(lyrics.isValid()) {
+    if(lyrics.isValid() && (!m_editingTrack || m_currentTrack.sameIdentityAs(*m_editingTrack))) {
```
ludouzi: Issue here is if the editor was open during a track change, the lyrics of the next track won't be displayed unless manually selected in the context menu.

lstrsrt: Pushed a fix. The `m_currentLyrics` assignment is now also behind a check like this so opening the lyrics editor on the next track won't show the previous lyrics anymore.

src/plugins/lyrics/lyricswidget.cpp
```diff
@@ -815,13 +823,18 @@ void LyricsWidget::openEditor(const Lyrics& lyrics)
...
+        dlg->saveState();
+        m_editingTrack = {};
+    });
     QObject::connect(dlg, &LyricsEditorDialog::lyricsEdited, this, [this](const Lyrics& updatedLyrics) {
         std::erase_if(m_lyrics, [](const Lyrics& existingLyrics) { return existingLyrics.isLocal; });
         if(updatedLyrics.isValid()) {
             m_lyrics.insert(m_lyrics.begin(), updatedLyrics);
         }
-        changeLyrics(updatedLyrics);
+        changeLyrics(updatedLyrics, true);
```
ludouzi, line 839: If the editor is open and the next track has no lyrics, this will clear the no lyrics display as `fullUpdate` will be false and `updatedLyrics` is valid.

We probably shouldn't be mutating `m_lyrics` here either if the tracks have changed while editing.

ludouzi (changes requested):

src/plugins/lyrics/lyricswidget.cpp
lstrsrt: Ok, that should be fixed now. Also gating `m_lyrics` updates now as suggested

ludouzi (approved):

ludouzi:
Thanks!

> By the way, is it intentional that the Lyrics tab in the properties dialog always uses the track's metadata lyrics?

Yeah, that is intentional for now. The other properties tabs operate only on data of the selected tracks, including artwork, so I kept the lyrics tab within the same scope. It can always be changed later if needed.

PR #1111, opened 2026-05-04
[filters] Add option to change cover source preference

### LINKED ISSUE #1021: Wrong Album Art…Sort Of
**Describe the bug**
When I have an album with some singles, I like to have the single’s artwork embedded in them. This is fine, and it shows up properly when playing. When a song is not a single or otherwise doesn’t have special artwork, I like to have the album cover embedded. This is also fine and also shows up properly.

However, when the first music file (seems to be based on track number, not alphabetical) uses single artwork instead of the album cover, that single artwork is what shows up for the entire album in fooyin.

If I go to Settings: Artwork and change Local Source to **Prefer directory artwork**, the albums show up correctly in the Library Filter, but singles with their own embedded artwork no longer display them while playing.

**To Reproduce**

1. Have multiple music files in one folder.
2. Have cover.jpg in the folder alongside the music files.
3. Embed the album cover in most of the files.
4. Embed single artworks in their specific songs.
5. Have the Local Source in Settings: Artwork be set to **Prefer embedded artwork**.
6. Have the first track be one of those singles.

**Expected behaviour**
I would really like it if fooyin could let me prefer “cover.jpg” for the Library Filter window but maintain embedded art for currently playing music.

**Screenshots**

Image

My layout showing two problem albums: A Flock of Seagulls’ self-titled and Adele’s 21.

Image

My directory for the AFOS album showing 1) the first track on the album, 2) the album cover, and 3) the first (alphabetical) file as an example of a song without its own single artwork.

Image

The second track playing, correctly showing the album cover.

Image

The view with **Prefer directory artwork** enabled. The Library Filter is now behaving correctly, but the currently playing track no longer shows its own artwork.

**System:**
 - OS: EndeavourOS
 - fooyin version: 0.10.3

**Additional context**
To be fair, I don’t think this is a fooyin problem specifically. However, since the dev seems to be very communicative, perhaps they can at least help me sort this out?

lstrsrt:
This adds a filters context menu option (Filter options > Display) to override the artwork local source preference from the global settings. I did this by first adding a `setSourcePreference` function to `CoverProvider` and then wiring it up to the filter model.

Additionally:
+ Fix a misspelled variable name.
+ Pass `CoverLoader` by reference in some `CoverProvider`-internal functions.

Closes #1021.

### DIFF
--- a/include/gui/coverprovider.h
+++ b/include/gui/coverprovider.h
@@ -27,6 +27,7 @@
 #include <QObject>
 
 #include <memory>
+#include <optional>
 #include <set>
 
 class QPixmap;
--- a/include/gui/coverprovider.h
+++ b/include/gui/coverprovider.h
@@ -37,6 +37,12 @@ namespace Fooyin {
 class AudioLoader;
 class SettingsManager;
 
+enum class ArtworkSourcePreference : uint8_t
+{
+    PreferDirectory = 0,
+    PreferEmbedded
+};
+
 /*!
  * Provides access to track album artwork.
  */
@@ -69,6 +75,13 @@ public:
      * @note this is enabled by default.
      */
     void setUsePlaceholder(bool enabled);
+
+    /*!
+     * Sets the local source preference for artwork.
+     * If std::nullopt is passed, the global user-facing setting will be used.
+     * @note the global setting is used by default.
+     */
+    void setSourcePreference(std::optional<ArtworkSourcePreference> preference);
 
     /** Returns @c true if @p track has a cover of the specific @p type. */
     [[nodiscard]] QFuture<bool> trackHasCover(const Track& track, Track::Cover type = Track::Cover::Front) const;
--- a/src/plugins/filters/filtermodel.cpp
+++ b/src/plugins/filters/filtermodel.cpp
@@ -158,6 +158,7 @@ public:
     CoverProvider::ThumbnailSize m_decorationSize;
     bool m_showLabels{true};
     Track::Cover m_coverType{Track::Cover::Front};
+    std::optional<ArtworkSourcePreference> m_coverSource{std::nullopt};
     std::vector<int> m_columnOrder;
 
     using ColumnAlignments = std::vector<Qt::Alignment>;
@@ -325,6 +326,11 @@ Track::Cover FilterModel::coverType() const
     return p->m_coverType;
 }
 
+std::optional<ArtworkSourcePreference> FilterModel::coverSource() const
+{
+    return p->m_coverSource;
+}
+
 void FilterModel::setRowHeight(int height)
 {
     p->m_rowHeight = height;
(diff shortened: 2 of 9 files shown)


include/gui/coverprovider.h
```diff
@@ -70,6 +76,13 @@ class FYGUI_EXPORT CoverProvider : public QObject
      */
     void setUsePlaceholder(bool enabled);
 
+    /*!
+     * Sets the local source preference for artwork.
+     * If std::nullopt is passed, the global user-facing setting will be used.
+     * @note the global setting is used by default.
+     */
+    void setSourcePreference(std::optional<ArtworkSourcePreference> preference);
```
ludouzi, line 85: `coverprovider.h` should include `#include <optional>` for this.

src/plugins/filters/filtermodel.cpp
```diff
@@ -388,6 +394,14 @@ void FilterModel::setCoverType(Track::Cover type)
     }
 }
 
+void FilterModel::setCoverSource(std::optional<ArtworkSourcePreference> source)
+{
+    if(std::exchange(p->m_coverSource, source) != source) {
+        p->m_coverProvider->setSourcePreference(source);
```
ludouzi, line 400: `FilterController` passes the same `CoverProvider` instance to every `FilterWidget`, so this will mutate the source preference for all filters, even if their own saved preference is different.

Each filter should probably store it's own instance of `CoverProvider`.

ludouzi (changes requested):

src/plugins/filters/filtermodel.cpp
lstrsrt: Sure, newest commit should address this

src/plugins/filters/filterwidget.h
```diff
@@ -151,6 +153,7 @@ class FilterWidget : public FyWidget
     ActionManager* m_actionManager;
     FilterColumnRegistry* m_columnRegistry;
     SettingsManager* m_settings;
+    CoverProvider* m_coverProvider;
```
ludouzi: This member isn't being used currently.

src/plugins/filters/filterwidget.cpp
```diff
@@ -89,15 +90,15 @@ class FilterView : public ExpandedTreeView
...
+                           LibraryManager* /*libraryManager*/, MusicLibrary* library,
+                           std::shared_ptr<AudioLoader> audioLoader, SettingsManager* settings, QWidget* parent)
     : FyWidget{parent}
     , m_actionManager{actionManager}
     , m_columnRegistry{columnRegistry}
     , m_settings{settings}
     , m_view{new FilterView(this)}
     , m_header{new AutoHeaderView(Qt::Horizontal, this)}
-    , m_model{new FilterModel(library, coverProvider, m_settings, this)}
+    , m_model{new FilterModel(library, new CoverProvider(std::move(audioLoader), settings), m_settings, this)}
```
ludouzi: `FilterModel` stores a raw `CoverProvider*`, so this will leak.

ludouzi (changes requested):

src/plugins/filters/filterwidget.cpp
lstrsrt: I went with setting the parent instead of changing it to a smart pointer.

ludouzi (approved):
Looks good, thanks!

PR #1130, opened 2026-05-07
[core] Several playback queue fixes

lstrsrt:
A couple of playback queue fixes.

* Removing tracks from the playback queue using the track context menu does not work anymore after a session restore. Reason: `PlaybackQueue::removeTracks()` uses a `set` of `PlaylistTrack`s which do not persist across sessions. Use a `set` of track hashes instead.
* Same issue exists for `PlaybackQueue::removeFirstMatchingTrack()`. I added two new `PlaybackQueue` functions that find a `PlaylistTrack` by hash, and fixed it by calling that instead.
* Double-clicking a track in a playlist or using the "Previous" action ignores whether that track is in the playback queue, so they won't be removed. Use these functions to fix that.

There is one more thing which I'm not sure is a bug:
Say we're at track X and then go to track N which is the only track in the queue. "Previous" will put us at track X-1 instead of N-1. However, "Next" will move us to N+1, not X+1. I don't know if this behavior is expected, so I left it alone for now.

### DIFF
--- a/src/core/playback/playbackqueue.cpp
+++ b/src/core/playback/playbackqueue.cpp
@@ -147,13 +147,13 @@ void PlaybackQueue::replaceTracks(const QueueTracks& tracks)
 
 std::optional<PlaylistTrack> PlaybackQueue::removeFirstMatchingTrack(const PlaylistTrack& track)
 {
-    const auto it = std::ranges::find(m_tracks, track);
-    if(it == m_tracks.end()) {
+    const auto index = getTrackIndex(track);
+    if(index == -1) {
         return {};
     }
 
-    PlaylistTrack removedTrack{*it};
-    m_tracks.erase(it);
+    PlaylistTrack removedTrack{m_tracks.at(index)};
+    m_tracks.erase(m_tracks.begin() + index);
     return removedTrack;
 }
 
--- a/src/core/playback/playbackqueue.cpp
+++ b/src/core/playback/playbackqueue.cpp
@@ -114,9 +114,9 @@ PlaylistTrack PlaybackQueue::nextTrackChange()
 
 int PlaybackQueue::getTrackIndex(const PlaylistTrack& track) const
 {
-    auto it = std::ranges::find_if(
-        m_tracks.cbegin(), m_tracks.cend(),
-        [hash = track.track.hash()](const PlaylistTrack& track) { return hash == track.track.hash(); });
+    const auto it = std::ranges::find_if(m_tracks, [&track](const PlaylistTrack& other) {
+        return track.playlistId == other.playlistId && track.sameIdentityAs(other);
+    });
 
     if(it == m_tracks.cend()) {
         return -1;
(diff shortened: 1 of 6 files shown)


ludouzi (commented):
`Track::hash` isn't ideal for this purpose. It's not a unique identity as multiple different tracks could share the same common metadata. It's now mainly used to share playback statistics between the same tracks with different file types e.g. flac and mp3.

Something like this would be more accurate:

```
bool isSameQueuedTrack(const PlaylistTrack& lhs, const PlaylistTrack& rhs)
{
    return lhs.track.sameIdentityAs(rhs.track);
}
```

Or we could add `sameIdentityAs` to `PlaylistTrack` and just delegate to the `Track` member.

lstrsrt:
My preferred way would be to keep the new `PlaybackQueue` functions but add `PlaylistTrack::sameIdentityAs()` and use it in `getTrackIndex()` and `removeTracks()`. I also noticed that we need to check if the playlist ids match (we could do it in `sameIdentityAs()`, but it might not always be wanted by callers).

src/core/playback/playbackqueue.cpp
```diff
@@ -112,6 +112,24 @@ PlaylistTrack PlaybackQueue::nextTrackChange()
     return track;
 }
 
+int PlaybackQueue::getTrackIndex(const PlaylistTrack& track) const
+{
+    const auto it = std::ranges::find_if(m_tracks, [&track](const PlaylistTrack& other) {
+        return track.playlistId == other.playlistId && track.sameIdentityAs(other);
```
ludouzi: Checking if the playlist ids match introduces a behaviour change if tracks are queued from a playlist but dequeued from a library viewer e.g. `Library Tree`.

Before, matching tracks could be dequeued from anywhere. Now, widgets like Library Tree show the `Remove from playback queue` action for queued tracks but it doesn't work, as they go through `dequeueTracks(const TrackList& tracks)`.

In that case, something like this would retain the previous behaviour while preventing tracks from being dequeued from different playlists (without a visible queue index):

```
bool isSameQueuedTrack(const PlaylistTrack& lhs, const PlaylistTrack& rhs)
  {
      if(!lhs.sameIdentityAs(rhs)) {
          return false;
      }

      return !lhs.playlistId.isValid()
          || !rhs.playlistId.isValid()
          || lhs.playlistId == rhs.playlistId;
  }
```
But the Playback Queue operates above all playlists, so maybe the user would expect to be able to dequeue tracks from anywhere, no matter the playlist.

lstrsrt: I see two possible fixes: either we introduce a setting that defaults to the previous behavior, or we always perform the previous behavior. To be honest I generally favor sane defaults over lots of settings + I doubt adding/removing tracks from different playlists is going to be a common user pattern (the playback queue doesn't even show the source playlist of a track anyway) so I will just restore the previous behavior. When per-playlist queues are added it might be a good setting to have though (so it can be possible to dequeue tracks globally for example).

Edit: I guess conditionally matching the playlist ID like you suggested would be a third option. For the reasons outlined above I am still more on the side of 2).

ludouzi: Agreed!

Also, the per-playlist queue on the ROADMAP was added years ago while the queue was still modeled exactly after fb2k. There haven't been any requests for such a feature, and with the recent additions and plans to have it optionally operate similar to MusicBee's queue, I'll probably remove it as a planned feature.

lstrsrt: I think we resolved it at the same time, lol.

lstrsrt:
Almost forgot: what about this?

>  Say we're at track X and then go to track N which is the only track in the queue. "Previous" will put us at track X-1 instead of N-1. However, "Next" will move us to N+1, not X+1. I don't know if this behavior is expected, so I left it alone for now.

ludouzi:
That's with `Follow last playback queue track` enabled, right? In that case, I'd say the more expected behaviour is to also apply it on previous as well, so `N-1`.

I'll commit a fix in a bit.

PR #1133, opened 2026-05-09
[core] Scripting: Fix $rand bounds

lstrsrt:
Fix an off-by-one bug in the base case & don't implicitly cast `uint` to `int`.


ludouzi (approved):
Thanks!

Issue #1129, opened 2026-05-07
Tag Reading: .mp3/ID3v2 multiple genres + .tak multiple values in any tag + .tak missing tags


ufomir:
**Describe the bug**

[TAG READING]
- .mp3: ID3v2 genre tag not read/interpreted properly if there are multiple values (returns one big semicolon-separated string), APEv2 tags read just fine
- .tak*: multi-value tags not read properly (reads just first the value)
- .tak*: if file is added via library scanning, DATE is missing. If file is added manually, DATE is shown in playlist view. In both cases DATE is **not** shown in Properties

*via ffmpeg + all support extensions option (.tak uses APEv2 tags)

**To Reproduce**
Zipped 3 example files used in screenshot and attached here: [report.zip](https://github.com/user-attachments/files/27471809/report.zip)

**Expected behaviour**
All tags present in supported files read and interpreted completely and correctly :)

**Screenshot**
<img width="912" height="448" alt="Image" src="https://github.com/user-attachments/assets/17924e2f-7d23-41fa-99a0-0fbeff70c61f" />Visible problems: "Rock;Pop;Punk;Classical Music" genre; Date shown in playlist but missing in Properties

**System:**
 - fooyin_0.10.5-trixie_amd64.deb

ufomir:
Just noticed the .tak related things might already be addressed in this [pull request #802](https://github.com/fooyin/fooyin/pull/802)

lstrsrt:
As you said, #802 should handle APE and .tak related bugs. #1137 should improve how we deal with string lists.

ludouzi:
https://github.com/fooyin/fooyin/pull/802 does handle APE and .tak, though it makes too many additional changes, some of which are now redundant, so I'll see if we can salvage anything from it.

On the ID3v2 MP3 parsing: That specific file uses a single value for the genre tag: 

<img width="396" height="245" alt="Image" src="https://github.com/user-attachments/assets/6552124f-f3fe-4753-9662-c44fae0d1ad0" />

fooyin (the TagLib reader at least) intentionally doesn't split those currently, as it tries to preserve the tags as they are in the file.We could add an option to control this, but I find it very helpful to find multi-value tags that were written incorrectly as a single value.

ufomir:
> We could add an option to control this, but I find it very helpful to find multi-value tags that were written incorrectly as a single value.

Had a closer look into this matter. I came from fb2k, which used to handle all my files and tags.

[...]

PR #802, opened 2026-01-02
Multivalue tag support, proper Apev2 support and better embedded cuesheet support

0spinboson:
##  Summary

  Adds proper multi-value tag support for APEv2-tagged files (TAK, WV, MPC, APE) and fixes several related bugs.

##  Features

  - Add $meta(field, index) function to retrieve a specific value from multi-value tags by index
    - Example: $meta(artist,0) returns only the first artist, avoiding cartesian product expansion
  - Add TAK to default supported extensions for FFmpeg reader
  - Add Cue_trackNN_* tag handling for embedded cuesheets
    - APEv2 files can store per-track metadata as Cue_track01_composer, Cue_track02_performer, etc.
    - These tags are now properly applied to the corresponding cue tracks
  - Add reload support for embedded cue tracks
    - "Reload tags from files" now works for tracks from embedded cuesheets

##  Bug Fixes

  - Fix multi-value tags (artists, genres, composers, etc.) being truncated for APEv2-tagged files
    - FFmpeg's metadata API truncates at null bytes; now read APEv2 tags directly
  - Fix duplicate tag values appearing after library rescan
  - Fix front/back cover art being confused in APEv2 files
  - Fix cue sheet parser overwriting file metadata (COMPOSER, GENRE, etc.) with empty values
  - Fix progress dialog not closing when all tracks are skipped during rescan

##  Tests

  - Add FFmpegReader tests for APEv2 tag reading, multi-value fields, and cover art

##  Test Plan

  - $meta(artist,0) returns only first artist
  - Multi-value tags display correctly for TAK/WV/MPC/APE files
  - Library rescan doesn't duplicate tag values
  - Front and back covers are correctly distinguished
  - Cue sheet files preserve metadata from audio file when not in cue
  - Embedded cue tracks with Cue_trackNN_* tags show correct per-track metadata
  - "Reload tags from files" works for embedded cue tracks
  - All tests pass

PR #1137, opened 2026-05-09
[core] Fix some FFmpeg tag decoding issues

lstrsrt:
Decoding tags with ffmpeg is not ideal at all for several reasons:

- There is no (or only selective) access to raw fields.
- There is no info about the tag type or version (like APE vs ID3v2.3 vs ID3v2.4).
- We get null-separated string lists as `char *` with no length field.
- and probably more.

This is a small attempt to fix some of these problems (see individual commit messages for details).
We can make an effort to get the tag type ourselves and at least split semicolon- or slash-separated string lists properly. This PR does not replace #802 which, from a cursory look, is better at handling APE tags.

Addresses #1129.


ludouzi (approved):
Thanks! Tested and working.

PR #1156, opened 2026-05-13
Replace cardinal directions with standard directions for tab stack

numerfolt:
Currently the tab position options of the tab stack widgets are named after the cardinal directions: north, east, south and west. And while this is consistent with the underlying code, it doesn't fit the program in my opinion.

This is why I suggest replacing the position options to standard directions according to the Qt documentations [description of the position enum](https://doc.qt.io/qt-6/qtabwidget.html#TabPosition-enum):
North → Above
East → Right
South → Below
West → Left


ludouzi (approved):
Agreed! Those are more natural for the UI.

PR #1168, opened 2026-05-14
[core,gui] Add isalpha, isalnum, isnum string functions

### LINKED ISSUE #982: Add alphabetic/alphanumeric/numeric string functions
My music library is organized in following way: `Music/{1, A, B, C, ...}/<artists>/<albums>`.
Now, I have a lengthy script for file ops, that automatically finds root folder for artists by stripping first character of artist name. However, there's a notable exclusion: artists which name starts from a non-alphabetic charactes. Few examples: `(ghost)`, `2 Unlimited`, `µ-Ziq`.
So I have actually to have two scripts: normal and special. Something that ideally should be unified with a bit more logic added on top. Hence, idea of scripting language extension with alphabetic/alphanumeric/numeric string functions.

Here's normal script, just in case:
```
$if($stricmp(%albumartist%,VA),VA,$upper($ascii($left($stripprefix(%albumartist%,the),1)))/$stripprefix(%albumartist%,the))/%albumartist% '('%date%')' %album%/%disctotal%-$num(%track%, 2). %artist% - %title%
```

Edit: Actually, intention is a bit wider to also to treat non-latin characters (greek, cyrillic, various hieroglyphs) as a single group that would go to that special folder.

P.S. Script Editor is FIRE!

bassstorm:
Add three new scripting functions to check character types:
- isalpha(): checks if string contains only alphabetic characters
- isalnum(): checks if string contains only alphanumeric characters
- isnum(): checks if string contains only numeric characters

These functions return ScriptResult allowing usage in conditionals. They support Unicode characters including Greek, Cyrillic, and other scripts.

Also adds reference entries in the GUI Script Editor for function discovery and autocomplete support.

Closes #982


ludouzi (approved):
Looks good, thanks!

PR #1170, opened 2026-05-14
[gui] PopupLineEdit: Don't emit editingFinished signal on escape

lstrsrt:
This is to fix edits to tab or playlist names still being accepted after pressing Escape because we erroneously emitted `editingFinished()` after `editingCancelled()`. Also, make `EditableTabBar::showEditor()` not accept empty text in inline editing to match the behavior of dialog edit mode.

### DIFF
--- a/src/gui/widgets/popuplineedit.cpp
+++ b/src/gui/widgets/popuplineedit.cpp
@@ -47,15 +47,17 @@ void PopupLineEdit::keyPressEvent(QKeyEvent* event)
     QLineEdit::keyPressEvent(event);
 
     if(event->key() == Qt::Key_Escape) {
+        m_cancelled = true;
         Q_EMIT editingCancelled();
     }
 }
 
 void PopupLineEdit::focusOutEvent(QFocusEvent* event)
 {
-    QLineEdit::focusOutEvent(event);
-
-    Q_EMIT editingFinished();
+    if(!m_cancelled) {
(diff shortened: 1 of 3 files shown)


ludouzi:
This changes focus loss behavior. Editing a tab and then clicking away without changing text leaves the popup editor visible but unfocused, because callers close it from `editingFinished()`.

lstrsrt:
Ah, I didn't notice because I only tested focus loss by clicking on other tabs (where it does actually close it).

src/gui/widgets/popuplineedit.cpp
```diff
@@ -47,15 +48,21 @@ void PopupLineEdit::keyPressEvent(QKeyEvent* event)
...
     if(event->key() == Qt::Key_Escape) {
+        m_cancelled = true;
         Q_EMIT editingCancelled();
     }
 }
 
 void PopupLineEdit::focusOutEvent(QFocusEvent* event)
 {
-    QLineEdit::focusOutEvent(event);
+    if(m_initialString == text()) {
```
lstrsrt, line 58: Could also listen to `textChanged()` and set a bool, but this covers if the text is first edited and then changed back to the original text.

ludouzi: Yeah, that is probably the better solution.

ludouzi (approved):
Thanks!

PR #1175, opened 2026-05-15
[core] Fix automatic rating scale detection for custom tags

bassstorm:
When reading custom rating tags with automatic scale mode enabled, the ratingTagFallback parameter was hardcoded to false, preventing automatic detection of rating scales (1-5, 1-10, etc.). This caused custom tag values to be incorrectly rejected as invalid.

The fix dynamically sets ratingTagFallback based on the configured rating scale, enabling automatic detection when RatingScale::Automatic is selected.


ludouzi (approved):
Looks good, thanks!

PR #1181, opened 2026-05-15
[core] Input: Improve artist slash check

lstrsrt:
Do `contains()` instead of an equality check so it works for multiple artists.


ludouzi (approved):
Thanks!

This will lead to artist strings like `AC/DC/Other` not splitting, but that's better for now than splitting into `AC`, `DC` and `Other` as before.

PR #1182, opened 2026-05-15
[core] Fix "Stop after this" when playback is stopped

lstrsrt:
Another quick fix: when playback is stopped and "Stop after this" is used, starting any track will set "Stop after current" instead. Reason is `m_playingIndex` being invalid in `playingTrackChanged()`. Verify it in the condition.


ludouzi (approved):
Thanks!

PR #1188, opened 2026-05-16
[core] Fix $ascii function accepting Latin1 Supplement characters

### LINKED ISSUE #1185: $ascii function exhibits unexpected behavior with Latin1 Supplement characters
**Describe the bug**
The function's purpose appears to be converting input strings to ASCII by:
* Normalizing to NFD form
* Extracting characters with code point < 128
* Attempting to convert higher Unicode points to their Latin1 equivalents

However, the current logic accepts any character that successfully converts to Latin1 (bytes 0-255), not just ASCII (0-127). This causes Latin1 Supplement characters (128-255) to be included in the output when they shouldn't be.

The problematic characters like µ (U+00B5), ¥ (U+00A5), and © (U+00A9) are in the Latin1-1 Supplement block and successfully convert to their Latin1 byte values, but they are not ASCII characters.

**To Reproduce**
Look at result of `$ascii(¥)` in Script Editor.

**Expected behaviour**
Empty for anything that cannot be converted to ASCII (0-127).

**Screenshots**
N/A

**System:**
 - OS: Debian 13
 - fooyin version: 0.10.7

**Additional context**
Initially I wanted to raise a PR fixing this, but wasn't sure about function behavior intent for wider audience.

bassstorm:
The $ascii function was incorrectly accepting characters from the Latin1 Supplement range (U+0080 to U+00FF) such as µ (micro sign), ¥ (yen sign), and © (copyright sign). These characters are not ASCII and should be excluded.

Fixes #1185

### DIFF
--- a/src/core/scripting/functions/stringfuncs.cpp
+++ b/src/core/scripting/functions/stringfuncs.cpp
@@ -151,12 +151,6 @@ QString ascii(const QStringList& vec)
         if(ch.unicode() < 128) {
             result.append(ch);
         }
-        else if(ch.category() != QChar::Mark_NonSpacing) {
-            const char asciiChar = ch.toLatin1();
-            if(asciiChar != 0 && static_cast<unsigned char>(asciiChar) < 128) {
-                result.append(QChar::fromLatin1(asciiChar));
-            }
-        }
     }
 
     return result;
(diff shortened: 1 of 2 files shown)


src/core/scripting/functions/stringfuncs.cpp
```diff
@@ -153,7 +153,7 @@ QString ascii(const QStringList& vec)
         }
         else if(ch.category() != QChar::Mark_NonSpacing) {
```
ludouzi: Any character reaching the `Latin1` check will be non-ASCII, so this branch is now redundant and can be removed.

bassstorm: Oh, you're right, only now I realized that was a double handling :) Fixed.

ludouzi (approved):

PR #1189, opened 2026-05-16
[core] Scripting: Show correct file size for selections and playlist

lstrsrt:
Currently, `%filesize%` and `%filesize_natural%` show the wrong file size for selections of more than 1 track or the whole playlist. Fix by extending `trackListValue()` to handle these variables & rename a couple related variables/functions to better reflect that they operate on track lists.

### DIFF
--- a/src/core/scripting/functions/tracklistfuncs.cpp
+++ b/src/core/scripting/functions/tracklistfuncs.cpp
@@ -41,7 +41,7 @@ QString playtime(const TrackList& tracks)
     return Utils::msToString(total);
 }
 
-QString playlistSize(const TrackList& tracks)
+uint64_t trackListSize(const TrackList& tracks)
 {
     uint64_t total{0};
     std::set<QString> seenSegmentPaths;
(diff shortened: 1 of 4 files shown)


ludouzi:
That is because `%filesize%` and `%filesize_natural%` were originally designed for a single track, while `%playlist_size%` was for playlists/lists.

This changes the format of `%playlist_size%`, which is unexpected:

image

This also breaks `%filesize%` and `%filesize_natural%` on a single-track subject, as it no longer returns the tracks metadata size due to checking `isTrackListVariableKind()` first:

image

ludouzi:
I'd also prefer to keep `%filesize%` and `%filesize_natural%` as single-track only for the time being, as they are documented and categorised as metadata variables, which are all single-track. 

Making them list-aware overloads the meaning based on evaluation context, which causes issues for any script environments which use fallbacks or placeholders like the playlist.

lstrsrt:
I see... in that case I'm not sure if it's worth making any changes at all. I'll close this.

Edit: Same timing again 👍️

PR #1196, opened 2026-05-17
[core] Reject non-finite ReplayGain values

lstrsrt:
One more edge case fix -- hope these aren't getting too much.
I discovered an album (ripped from Tidal) in my library yesterday that had "NaN dB" set for `REPLAYGAIN_ALBUM_GAIN`. This made me realize that we currently accept non-finite RG values. There is a sanity check in the RG processor, but it doesn't catch `+inf` or `-inf` and it seems simpler to treat the values as invalid from the start. Fix: Add `isfinite()` checks to the metadata readers and `Track` RG predicates.

(There are some other `QString::toFloat()` calls in the codebase that might benefit from the same treatment.)


ludouzi (approved):
Thanks!

Yeah, there are a few other places which could use similar guards. I'll handle a few of them in a follow up.

PR #1206, opened 2026-05-19
[gui] SelectionInfo: Show tooltips when text is elided

lstrsrt:
I think it makes sense for `SelectionInfo` to show tooltips if needed. Mostly useful when the horizontal scrollbar is disabled and you still want to easily see full file paths etc.


ludouzi:
Great idea. This should be added to the second constructor used for the properties tab as well.

lstrsrt:
Done

ludouzi (approved):
Looks good!

PR #1222, opened 2026-05-25
[gui] Expose selected playlist to GUI plugins

piotrwicijowski:
## Summary

Expose the currently selected playlist in the main playlist UI to GUI plugins through `GuiPluginContext`.
This adds a small GUI-plugin-facing controller:

- `CurrentPlaylistController`
- `currentPlaylist()`
- `currentPlaylistId()`
- `changeCurrentPlaylist(const UId&)`
- `currentPlaylistChanged(previous, current)`

## Motivation

The [fooyin-vim-plugin](https://github.com/piotrwicijowski/fooyin-vim-plugin) needs a reliable way to work with the playlist the user is currently interacting with.
Read access is needed to target the selected playlist correctly.
Write access is needed for follow-up features such as:

- switching to next / previous playlists
- restoring cross-playlist marks
- any plugin workflow that needs to move the UI to a specific playlist before restoring focus inside it
At the moment that state is not exposed through the GUI plugin API, so plugin code has to infer it indirectly. This change makes the selected playlist available directly, and also allows plugins to switch it, without changing existing playback-oriented semantics.

## Behaviour

`CurrentPlaylistController` follows the playlist selected in the main playlist UI and is intentionally separate from the active playback playlist.
It:
- reports `nullptr` / invalid id when no playlist is selected
- emits when the selected playlist changes
- does not emit when only the active playback playlist changes
- does not emit when reselecting the same playlist
- allows switching the selected playlist by id
- ignores invalid ids safely
- performs a no-op when asked to switch to the already selected playlist

## Implementation

- add public `CurrentPlaylistController` under `include/gui/playlist/`
- wire a concrete controller implementation to `PlaylistController`
- expose it through `GuiPluginContext`
- route `changeCurrentPlaylist(const UId&)` through the existing selected-playlist controller path
- add GUI tests covering selection identity, mutation, and signal behaviour

## Testing

Commands used:

```bash
cmake -B build-agent -G Ninja -DBUILD_TRANSLATIONS=OFF
cmake --build build-agent
cmake -B build-test-agent -DBUILD_TESTING=ON -DBUILD_TRANSLATIONS=OFF -G Ninja
cmake --build build-test-agent
QT_QPA_PLATFORM=offscreen /fooyin/build-test-agent/run/bin/test_currentplaylistcontroller
```


piotrwicijowski:
FreeBSD failure appears unrelated to this change and is most likely caused by the current Clang + ccache + PCH CI setup

piotrwicijowski:
Updated the PR to be able to change the current playlist as well

ludouzi:
Thanks for the PR.

This seems a little overengineered. `CurrentPlaylistControllerImpl` doesn't add anything beyond forwarding to `PlaylistController`, and the added tests are really just testing `PlaylistController` instead of `CurrentPlaylistController`. 

I’d prefer this as a small concrete `CurrentPlaylistController` owned by `GuiApplication`, without the unnecessary inheritance.

piotrwicijowski:
Hi,
Yeah, I get it, coming from a more corporate dayjob where generous layers of indirection is the default, I felt that it would be ok here.
The new version should be much simpler.
Full disclosure - the code in this PR is heavily slopped, but with close oversight.

ludouzi:
Thanks!

I took another look and realised this can be simplified further. Since `CurrentPlaylistController` is just exposing a narrow subset of `PlaylistController`, I made it an interface again which `PlaylistController` implements directly. That also removes the callbacks and extra signal forwarding.

piotrwicijowski:
Thank you so much for your help and very quick turnaround, this addition will make the vim plugin much more capable. I'm planning a couple of more features that will require some minimal plugin API changes soon, so more PRs will come soon.

PR #1244, opened 2026-05-31
gui: add plugin script variables and playlist view refresh API

piotrwicijowski:
## Summary

This PR adds two public GUI plugin extension points needed by the vim plugin's `%mark%` and `%global_mark%` support:

1. plugin-defined script variables
2. plugin-triggered playlist row refresh for transient rendered state

That lets GUI plugins expose variables to fooyin's scripting surfaces and refresh affected playlist rows when the underlying plugin-owned state changes.

### 1. Public plugin script-variable registration

Adds a public `ScriptVariableRegistry` to `GuiPluginContext`.

GUI plugins can register `ScriptVariableProvider`s through public API rather than depending on internal GUI types.

### 2. Global script-variable support in the script engine

`ScriptParser` now supports global providers, and `ScriptRegistry` invalidates cached bindings when that provider set changes.

That makes plugin-defined variables available anywhere scripts are parsed and rebound, including playlist columns and script editor usage.

### 3. Script reference integration

Plugin-defined variable metadata is included in the script reference entries so the variables show up in the scripting UI with their category and description.

### 4. Public playlist refresh hook for transient plugin state

Adds a public `PlaylistViewRefresher` to `GuiPluginContext`.

Plugins can request:

- refresh of a playlist by `playlistId`
- targeted refresh of specific entries by `playlistId` and `entryId`

This is for plugin-owned transient state that affects rendered playlist scripts without going through fooyin's metadata update flow.

## Why this is needed

The [vim plugin](https://github.com/piotrwicijowski/fooyin-vim-plugin) stores marks as plugin-owned state keyed by playlist and entry identity.

To expose mark state in playlist script evaluation and keep rendered columns in sync when marks change, it needs:

1. a public way to register script variables
2. a public way to invalidate the affected playlist rows

This PR provides both in generic form for GUI plugins.

## API surface

New public types:

- `Fooyin::ScriptVariableRegistry`
- `Fooyin::PlaylistViewRefresher`

New `GuiPluginContext` fields:

- `scriptVariableRegistry`
- `playlistViewRefresher`

## Testing

Added or updated coverage for:

- global script variable provider registration in `tests/core/scriptparsertest.cpp`
- script reference exposure for plugin variables in `tests/gui/scriptreferenceentriestest.cpp`

## Notes

This is intentionally generic GUI/plugin infrastructure rather than a vim-specific API.

These changes were created with help of AI.


piotrwicijowski:
@ludouzi - sorry to bother you, but do you foresee the changes from this PR to be valuable to fooyin? The PR has been in a limbo for some time now, and I'm wondering if it is worth trying to rebase it.

ludouzi:
I'll go through it a bit later, but my main concern is the mutex added to the main evaluation hot path. A 10,000 track playlist across several columns could perform hundreds of thousands of mutex operations, which is likely to lead to performance regressions.

piotrwicijowski:
@ludouzi - yup, agreed, the mutex in the hot path is an overkill. The registration of the script variables is now moved into the startup (assuming that for now we don't need plugins to dynamically register variables during runtime), so the locking should be much less aggressive.

piotrwicijowski:
@ludouzi gentle ping.

PR #1246, opened 2026-05-31
[spectrum] Fix static layer DPR source rect

lstrsrt:
Same idea as #1075, fixes rendering on high-DPI displays. I tried modeling it after the last fix :)


ludouzi (approved):
Thanks - knew I forgot something!

There are also issues with band/bar spacing in `Notes` mode, so I'll make a follow-up fix for that too.

PR #1247, opened 2026-05-31
[gui] Playlist: Scale background pixmap with DPR

lstrsrt:
This should fix the last high-DPI issue I know of, affecting background images in playlists.

### DIFF
--- a/src/gui/playlist/playlistview.cpp
+++ b/src/gui/playlist/playlistview.cpp
@@ -69,7 +69,7 @@ QPixmap blurredPixmap(const QPixmap& pixmap, qreal radius)
     }
 
     QImage blurred = pixmap.toImage().convertToFormat(QImage::Format_ARGB32_Premultiplied);
-    qt_blurImage(blurred, radius, true, false);
+    qt_blurImage(blurred, radius, true);
     return QPixmap::fromImage(blurred);
 }
 } // namespace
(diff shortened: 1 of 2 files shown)


src/gui/playlist/playlistview.cpp
```diff
@@ -542,10 +544,15 @@ QPixmap PlaylistView::preparedBackgroundPixmap(const QPixmap& source, Background
...
+    targetSize = {static_cast<int>(std::ceil(static_cast<qreal>(targetSize.width()) * dpr)),
+                  static_cast<int>(std::ceil(static_cast<qreal>(targetSize.height()) * dpr))};
+
     cache.sourceKey    = source.cacheKey();
     cache.viewportSize = viewportSize;
+    cache.dpr          = dpr;
     cache.pixmap       = source.scaled(targetSize, Qt::IgnoreAspectRatio, Qt::SmoothTransformation);
-    cache.pixmap       = blurredPixmap(cache.pixmap, m_bgOptions.blur);
+    cache.pixmap.setDevicePixelRatio(dpr);
+    cache.pixmap = blurredPixmap(cache.pixmap, m_bgOptions.blur);
```
lstrsrt: It might be desirable to scale the blur pixels with the DPR as well.

ludouzi: Yeah, it would be ideal to scale the blur radius as well. Otherwise, a configured `20px` blur becomes `~10px` at 2x.

lstrsrt: Done. I had to push it again because I canceled the workflow on accident.

ludouzi (approved):
Thanks!

PR #1277, opened 2026-06-09
fix: migrate TrackStats.Rating to REAL to match stored values

### LINKED ISSUE #1216: [Core/Database] type mismatch for TrackStats.Rating: schema says INTEGER, but fooyin inserts in float
### **Describe the bug**
[dbschema.xml](https://github.com/fooyin/fooyin/blob/master/data/dbschema.xml) defines
`Rating INTEGER DEFAULT 0` (no revisions)
but updates as float in [trackdatabase.cpp](https://github.com/fooyin/fooyin/blob/master/src/core/database/trackdatabase.cpp)

### **To Reproduce & Screenshots**
Open fooyin.db in DB Browser SQLite
On tab "Database Structure" see TrackStats definition
Image
On tab Browse Data" see Rating values as float
Image

### **Expected behaviour**
Matching schema and data. Schema should be updated.
Curious, why fooyin's sqlite engine does not complain, though. :0

### **System:**
 - fooyin version: 0.10.8 Trixie

### **Additional context**
Tried to write a python script to import track statistics from foobar (allows export to xml) to fooyin.

mvanhorn:
## Summary
Adds DB schema revision 18 that changes the `TrackStats.Rating` column affinity from `INTEGER` to `REAL`, so the declared schema type matches the floating-point rating values fooyin actually stores.

## Why this matters
Resolves #1216. `trackdatabase.cpp` reads and writes ratings as `float` (`normaliseTrackRating(float)`, `value(...).toFloat()`, float binds), but the schema declared `Rating INTEGER DEFAULT 0`. SQLite's dynamic typing means no runtime error occurs, but external tooling that inspects the declared schema (the reporter was importing foobar2000 statistics via a script) sees the wrong type. The maintainer (@ludouzi) confirmed the mismatch and said a migration should be added.

## Changes
- `data/dbschema.xml`: new `<revision version="18">` retyping `TrackStats.Rating` to `REAL`, following the established rename-old / add-new / copy / drop-old pattern used by the revision 12 ReplayGain retype (SQLite cannot alter a column type in place).
- `src/core/database/database.cpp`: bump `CurrentSchemaVersion` from 17 to 18 so the upgrade engine (`upgradeDatabase` loops `while(nextVersion < targetVersion)`) actually applies the new revision.

## Testing
- `xmllint --noout data/dbschema.xml` passes.
- Verified existing rating read/write paths in `trackdatabase.cpp` round-trip values as `float` and do not depend on the INTEGER affinity, so existing data is preserved across the migration.
- Note: a full Qt build was not run locally; the change is a data-schema revision plus a one-line version constant bump following an existing, exercised migration pattern.

Fixes #1216

AI was used for assistance.

### DIFF
--- a/data/dbschema.xml
+++ b/data/dbschema.xml
@@ -378,4 +378,18 @@
             ALTER TABLE Tracks ADD COLUMN CreatedDate INTEGER;
         </sql>
     </revision>
+    <revision version="18">
+        <description>
+            Change TrackStats.Rating column affinity to REAL.
+        </description>
+        <sql>
+            ALTER TABLE TrackStats RENAME COLUMN Rating TO RatingOld;
+
+            ALTER TABLE TrackStats ADD COLUMN Rating REAL DEFAULT 0;
+
+            UPDATE TrackStats SET Rating = RatingOld;
+
+            ALTER TABLE TrackStats DROP COLUMN RatingOld;
+        </sql>
+    </revision>
 </schema>
--- a/src/core/database/database.cpp
+++ b/src/core/database/database.cpp
@@ -30,7 +30,7 @@
 
 using namespace Qt::StringLiterals;
 
-constexpr auto CurrentSchemaVersion = 17;
+constexpr auto CurrentSchemaVersion = 18;
 
 namespace {
 Fooyin::DbConnection::DbParams dbConnectionParams()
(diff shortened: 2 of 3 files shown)


ludouzi (approved):

ludouzi:
Thanks. Tested and working.

mvanhorn:
Appreciate the merge, @ludouzi. TrackStats.Rating stored as REAL ends the truncated-ratings mismatch.

PR #1281, opened 2026-06-11
[gui] Save window size on layout switch

lstrsrt:
Since #1267, layout switching saves the previous layout. We were already saving colors and fonts, but not the window size. This meant that switching from a layout would lose its window size, even if the user had previously manually saved it via the export dialog.

Fix: Allow `EditableLayout::saveCurrentToLayout()` to save the window size, controlled with an argument.

Note: I've been seeing a bug where switching layouts sometimes causes the current playlist to not be re-rendered/showing up empty (fixed by switching to another playlist and then back). For some reason, applying this patch makes it much rarer. Not sure what the relation is though!

### DIFF
--- a/src/gui/editablelayout.cpp
+++ b/src/gui/editablelayout.cpp
@@ -715,10 +715,10 @@ FyLayout EditableLayout::saveCurrentToLayout(const QString& name, bool saveWindo
 
     const auto theme = p->m_settings->value<Settings::Gui::Theme>().value<FyTheme>();
     if(theme.isValid()) {
-        if(saveWindowSize) {
-            layout.saveWindowSize();
-        }
         layout.saveTheme(theme);
+    }
+    if(saveWindowSize) {
+        layout.saveWindowSize();
     }
 
     return layout;
--- a/src/gui/editablelayout.cpp
+++ b/src/gui/editablelayout.cpp
@@ -688,7 +688,7 @@ void EditableLayout::initialise()
     p->changeLayout(p->m_layoutProvider->currentLayout());
 }
 
-FyLayout EditableLayout::saveCurrentToLayout(const QString& name)
+FyLayout EditableLayout::saveCurrentToLayout(const QString& name, bool saveWindowSize)
 {
     QJsonObject root;
     QJsonArray array;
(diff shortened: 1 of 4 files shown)


src/gui/layoutcommands.cpp
```diff
@@ -60,20 +60,20 @@ bool LayoutChangeCommand::checkContainer()
 SwitchLayoutCommand::SwitchLayoutCommand(EditableLayoutPrivate* editableLayout, FyLayout layout)
     : LayoutChangeCommand{editableLayout->m_self}
     , m_editableLayout{editableLayout}
-    , m_oldLayout{m_layout->saveCurrentToLayout(editableLayout->m_layoutProvider->currentLayout().name())}
+    , m_oldLayout{m_layout->saveCurrentToLayout(editableLayout->m_layoutProvider->currentLayout().name(), true)}
```
lstrsrt, line 63: We could also only make it save the window size if it's already defined in the layout. But if the user has just changed the window size manually, they probably expect it to be saved as well.

src/gui/editablelayout.cpp
```diff
@@ -715,6 +715,9 @@ FyLayout EditableLayout::saveCurrentToLayout(const QString& name)
 
     const auto theme = p->m_settings->value<Settings::Gui::Theme>().value<FyTheme>();
     if(theme.isValid()) {
+        if(saveWindowSize) {
+            layout.saveWindowSize();
+        }
```
ludouzi: This should be moved outside the theme block as otherwise it's dependent on a valid theme which isn't the case when all colours and fonts are at system defaults.

src/gui/layoutcommands.cpp
ludouzi: Agreed. The theme is saved if valid, so it makes sense to also save the window size as well.

src/gui/editablelayout.cpp
lstrsrt: Fixed

ludouzi (approved):

ludouzi:
Thanks!

The empty playlist issue sounds like a race condition. Maybe the small amount of extra work in retrieving and saving the window size changes event-loop timing, potentially masking the issue. I'll take a look into it.

PR #1283, opened 2026-06-12
[gui] Playlist: Only set font from resolved app style

lstrsrt:
`PlaylistWidget` currently calls `setFont()` on its model in two places: `applyInitialViewSettings()` and `refreshViewStyle()`. This causes a race where the initial/system font sometimes wins out over a custom Default font (almost always on optimized builds and only on startup). Fix by removing the first call and letting `refreshViewStyle()` handle it.

### DIFF
--- a/src/gui/playlist/playlistwidget.cpp
+++ b/src/gui/playlist/playlistwidget.cpp
@@ -277,6 +277,8 @@ PlaylistWidget::PlaylistWidget(ActionManager* actionManager, PlaylistInteractor*
     applyInitialViewSettings();
     applyBackgroundSettings();
 
+    refreshViewStyle();
+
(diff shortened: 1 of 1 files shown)


ludouzi:
That results in newly created playlists not fully following the configured font:

image

We'll probably want to call `refreshViewStyle` somewhere during construction, and guard against the initial default value using `ResolvedAppStyle::revision` > 0.

lstrsrt:
I couldn't get this to happen for me but I think the newest commit guards against it. Also just saw this bug happen with the library tree while testing, so that will need a similar change.

ludouzi (approved):

ludouzi:
Thanks. A few other widgets as well as the Library Tree will also need to be handled, so I'll see if this fix can be generalised.

PR #1284, opened 2026-06-12
feat: add independent sort field to filter column settings

### LINKED ISSUE #1279: Add sort field to filter column settings
Please add a sort field to filter columns settings so that we can sort lines in the filter widgets independently of displayed text.

For example, my use case is for sorting albums released in the same year while displaying only year. So display script can be `%year% - %album%` and in sort script i can use `%date% - %album%`

Quick mock-up:
Image

mvanhorn:
## Summary

Filter columns get an optional sort field, so rows can be sorted by a script independent of the displayed text. The motivating case from #1279: show only `%year%` but sort albums from the same year by `%date% - %album%`. When the sort field is empty, sorting falls back to the displayed `field`, so existing filters are unchanged.

## Why this matters

@ludouzi confirmed the hidden-sort-column workaround works but agreed a dedicated sort field is cleaner, and the reporter noted the workaround re-sorts randomly when switching artists. The new field is wired through the same paths the displayed columns use, so it persists and round-trips like the rest of the column config.

## Changes

- `FilterColumn` gains a `sortField` QString with `operator==` and `QDataStream` read/write updated; an empty default keeps backward compatibility with existing saved layouts.
- `filterrows.cpp` evaluates `sortField` (falling back to `field`) into a per-column sort string on the `FilterItem`, and `FilterSortModel::lessThan` compares those sort strings instead of the displayed text.
- The column editor (`filterscolumnmodel`) exposes the sort script as an editable column in settings.

## Testing

Built locally; verified the sort field persists across restart, falls back to the displayed field when empty, and re-sorts correctly when switching the active filter selection.

Fixes #1279

### DIFF
--- a/src/plugins/filters/filtercolumneditordialog.cpp
+++ b/src/plugins/filters/filtercolumneditordialog.cpp
@@ -64,8 +64,9 @@ FilterColumnEditorWidget::FilterColumnEditorWidget(FilterColumnRegistry* columnR
     m_columnList->hideColumn(0);
     m_columnList->setExtendableColumn(1);
     m_columnList->verticalHeader()->hide();
-    m_columnList->horizontalHeader()->setStretchLastSection(true);
+    m_columnList->horizontalHeader()->setStretchLastSection(false);
     m_columnList->horizontalHeader()->setSectionResizeMode(QHeaderView::ResizeToContents);
+    m_columnList->horizontalHeader()->setSectionResizeMode(2, QHeaderView::Stretch);
 
     m_openEditor->setText(tr("Script Editor"));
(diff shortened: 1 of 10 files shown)


ludouzi:
Thanks. I made a few changes: 

- Sort values are now stored explicitly instead of in rich text
- Stale sort values are avoided on row updates
- UI shows empty sort fields as `Use display script`, which matches similar widget column pages

ludouzi (approved):

mvanhorn:
Appreciate you merging this, @ludouzi. An independent sort field on filter columns helps when display and sort order differ.

PR #1285, opened 2026-06-12
[plugins] FileOps: Make destination a dropdown with library roots

bassstorm:
Another quality-of-life improvement for FileOps dialog. This change makes destination field a dropdown that contains library root paths for quick preselection. Existing functionality to call file chooser is preserved.

Here's an example of the updated UI for my library consisting of two library roots: `/media/cdrom` and `/media/music`.

image

The feature implementation appeared to be a bit more tricky than I anticipated. But I gave it a fair share of testing and can see no obvious issues from user perspective. Feedback is appreciated as always.

### DIFF
--- a/src/plugins/fileops/fileopsdialog.cpp
+++ b/src/plugins/fileops/fileopsdialog.cpp
@@ -44,6 +44,7 @@
 #include <QPushButton>
 #include <QRadioButton>
 #include <QTreeView>
+#include <core/library/librarymanager.h>
 
 using namespace Qt::StringLiterals;
 
@@ -56,7 +57,8 @@ class FileOpsDialogPrivate : public QObject
 
 public:
     FileOpsDialogPrivate(FileOpsDialog* self, MusicLibrary* library, const TrackList& tracks, Operation operation,
-                         std::shared_ptr<AudioLoader> audioLoader, SettingsManager* settings);
+                         std::shared_ptr<AudioLoader> audioLoader, SettingsManager* settings,
+                         LibraryManager* libraryManager);
 
     void setup();
 
@@ -84,6 +86,8 @@ public:
 
     FileOpsDialog* m_self;
     SettingsManager* m_settings;
+    MusicLibrary* m_library;
+    LibraryManager* m_libraryManager{nullptr};
 
     Operation m_operation;
 
@@ -92,7 +96,7 @@ public:
     QRadioButton* m_renameOp;
     QRadioButton* m_extractOp;
 
-    QLineEdit* m_destination;
+    QComboBox* m_destination;
     ScriptLineEdit* m_filename;
     QCheckBox* m_entireSource;
     QCheckBox* m_removeEmpty;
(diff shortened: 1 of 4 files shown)


src/plugins/fileops/fileopsdialog.cpp
```diff
@@ -415,6 +431,25 @@ void FileOpsDialogPrivate::populatePresets()
...
+    // Populate with library roots
+    if(m_libraryManager && m_libraryManager->hasLibrary()) {
+        const auto& libs = m_libraryManager->allLibraries();
+        for(const auto& entry : libs) {
+            const QString path = entry.second.path;
+            if(path.isEmpty()) {
+                continue;
+            }
+            if(m_destination->findText(path) == -1) {
+                m_destination->addItem(path, path);
```
ludouzi: This should just be `m_destination->addItem(path);`

src/plugins/fileops/fileopsdialog.cpp
```diff
@@ -297,8 +302,18 @@ void FileOpsDialogPrivate::updateButtonState() const
...
+    const QString text  = m_destination->lineEdit() ? m_destination->lineEdit()->text() : m_destination->currentText();
+    const int index     = m_destination->currentIndex();
+    const QVariant data = index >= 0 ? m_destination->itemData(index) : QVariant{};
+
+    if(index >= 0 && data.isValid() && !data.toString().isEmpty() && text == m_destination->itemText(index)) {
+        preset.dest = data.toString();
+    }
+    else {
+        preset.dest = text;
+    }
```
ludouzi: All of this can be reduced to `preset.dest = m_destination->currentText();`

src/plugins/fileops/fileopsdialog.cpp
```diff
@@ -474,11 +509,15 @@ void FileOpsDialogPrivate::modelUpdated()
 
 void FileOpsDialogPrivate::browseDestination() const
 {
-    const QString path = !m_destination->text().isEmpty() ? m_destination->text() : QDir::homePath();
+    const QVariant data = m_destination->currentData();
+    const QString start
+        = (data.isValid() && !data.toString().isEmpty()) ? data.toString() : m_destination->currentText();
+    const QString path = !start.isEmpty() ? start : QDir::homePath();
```
ludouzi: This can be simplified to:

```
const QString start = m_destination->currentText();
const QString path = !start.isEmpty() ? start : QDir::homePath();
```

ludouzi (changes requested):

src/plugins/fileops/fileopsdialog.cpp
bassstorm: Oh true, good spot! That's leftover from original idea to have dropdown entries like "Root Name (/root/path)", that appeared to be more confusing than helping, hence discarded.

ludouzi (approved):
Thanks!

PR #1300, opened 2026-06-16
[gui] Fix library folder picker on macOS

redanthrax:
Native NSOpenPanel returned no path when adding a library from the modal settings dialog, so the pending row never appeared. Use a single Qt directory dialog on macOS, hide settings while picking, and defer the prompt. Set MACOSX_BUNDLE_GUI_IDENTIFIER so native file dialogs work in release bundles.


ludouzi:
This was also an issue on Linux and Windows due to recent changes in Qt. I can't reproduce it on any system after https://github.com/fooyin/fooyin/commit/037af36b41b6ddbcf903726abe62e8e48f4d9a70. Does that not resolve it on your macOS machine?

PR #1302, opened 2026-06-17
[gui] Try to show active theme in theme settings

lstrsrt:
Instead of always showing `System default`, try to display the actual theme name when loading theme settings.

### DIFF
--- a/src/gui/settings/guithemespage.cpp
+++ b/src/gui/settings/guithemespage.cpp
@@ -183,6 +183,16 @@ void GuiColoursPageWidget::load()
 
     loadDefaults();
     loadCurrentTheme();
+
+    const auto currentTheme = this->currentTheme();
+    for(const auto& theme : themes) {
+        if(theme.colours == currentTheme.colours && theme.fonts == currentTheme.fonts) {
+            if(const int index = m_themesBox->findText(theme.name); index >= 0) {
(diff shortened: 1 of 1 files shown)


src/gui/settings/guithemespage.cpp
```diff
@@ -183,6 +183,16 @@ void GuiColoursPageWidget::load()
 
     loadDefaults();
     loadCurrentTheme();
+
+    const auto currentTheme = this->currentTheme();
+    for(const auto& theme : themes) {
+        if(theme.colours == currentTheme.colours && theme.fonts == currentTheme.fonts) {
```
lstrsrt: This whole loop is a hack, it would be a lot smarter to get the current theme name directly. I didn't find any function that returns it though (the `CustomTheme` setting doesn't seem to store it).

ludouzi: `CustomTheme` was only [just](https://github.com/fooyin/fooyin/commit/82b27e62d32e8682a9cc8efa977f74baa7c21990#diff-72769e9509dbcef538b20ad9a49ae0c9bcac537fdc698dea69c75455865a3c53) changed to a persistent setting. Quick Setup correctly sets this, but not the theme settings page yet. I'll make some changes.

ludouzi (approved):

Issue #1152, opened 2026-05-12
Scrobbling minimum threshold


bassstorm:
Not sure if that's a feature or a bug.
I set played threshold to 100%. I guess, it's used to decide if a track is to be scrobbled as well. Now, there's albums with last track of 20+ minutes that includes a hidden track after a long silence. Usually I fast forward the silence to reach additional audio content and finish the track. In Foobar2000 such tracks are scrobbled no issue, but with Fooyin that wasn't the case.
So, by the look of it I guess there's no such a thing as scrobbling threshold (as Last.fm suggests 4 minutes or 50% of track duration, whichever comes first). Can you have a look please?

P.S. I understand setting played threshold to 50% is a viable workaround, but scrobbling should work for all possible settings, since even 50% doesn't guarantee that audio content is longer than silence. For example, https://www.discogs.com/release/2759-Autechre-LP5.

ludouzi:
Yeah, it's currently based on the played threshold.

It'd probably be best to follow the Last.fm suggestion.

PR #1307, opened 2026-06-19
[core,gui,scrobbler] Introduce dual playback threshold

bassstorm:
This PR proposes a better (imo) approach to tackle #1152. Inspired by Pano Scrobbler, it adds a second, time-based threshold, so that playcount threshold is uniformerly managed across all components. The first threshold reached wins.

Min: 30 seconds.
Default: 4 minutes.
Max: 10 minutes.

image

P.S. Not sure if extra flexibility like disabling a threshold is really necessary. But if so, just say the word.


ludouzi (approved):
Looks good, thanks!

PR #1308, opened 2026-06-19
[sleepinhibitor] Inhibit logout and pass valid app ID

lstrsrt:
I just learned that GNOME can actually show a warning in the logout/shutdown/reboot confirmation screen when inhibitors are active, including the name and reason of the running app. However, this only works when we 1) add the logout block flag and 2) pass a valid app identifier to `Inhibit()`. So let's do that, and also make the reason translatable since it is now visible to users.


ludouzi (approved):
Thanks!

PR #1312, opened 2026-06-20
[core] Scripting: Add bounds check to $chop()

lstrsrt:
This patch fixes undefined behavior when `$chop()` is called with `count` exceeding the length of `text`.


ludouzi (approved):
Thanks!

PR #1313, opened 2026-06-20
[gui] Preserve scroll position upon view state change

### LINKED ISSUE #1221: Scroll position is reset upon changing details that are part of playlist preset
**Describe the bug**
I use a customized Album/Disc preset. One of the changes is a subheader to display disc subtitle if present:
```
$ifgreater(%disctotal%,1,Disc #%disc%)[ ― %discsubtitle%]
```

Now, when I add a disc subtitle to a bespoke album in a long playlist, every time I end up at the very top of playlist once update is applied.

**To Reproduce**
1. Have a playlist of few dozens of albums.
2. Change some details like described above. Sometimes even album artist change is enough.
3. See scroll position is reset as playlist is re-rendered.

**Expected behaviour**
Scroll position stays intact.

**Screenshots**
Easy to reproduce, but I can provide a screencast if need be.

**System:**
 - OS: Debian 13, Flatpak
 - fooyin version: 0.10.8

**Additional context**
N/A

bassstorm:
Fixes #1221

### DIFF
--- a/src/gui/playlist/editableplaylistsession.cpp
+++ b/src/gui/playlist/editableplaylistsession.cpp
@@ -794,19 +794,17 @@ void EditablePlaylistSession::handleTracksChanged(PlaylistWidgetSessionHost& ses
                   return;
               }
 
-              if(allNew) {
+              QModelIndex topIndex;
+              if(currentPlaylist->trackCount() > 0 && viewState.topIndex >= 0) {
+                  topIndex = host.playlistModel()->indexAtPlaylistIndex(viewState.topIndex, false);
+              }
+
+              if(allNew && !topIndex.isValid()) {
                   host.playlistView()->playlistReset();
                   host.playlistView()->scrollToTop();
                   return;
               }
 
-              if(currentPlaylist->trackCount() > 0 && viewState.topIndex >= 0) {
-                  const QModelIndex topIndex = host.playlistModel()->indexAtPlaylistIndex(viewState.topIndex, false);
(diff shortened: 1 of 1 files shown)


src/gui/playlist/editableplaylistsession.cpp
```diff
@@ -794,19 +794,17 @@ void EditablePlaylistSession::handleTracksChanged(PlaylistWidgetSessionHost& ses
                   return;
               }
 
-              if(allNew) {
+              QModelIndex topIndex;
+              if(currentPlaylist->trackCount() > 0 && viewState.topIndex >= 0) {
+                  topIndex = host.playlistModel()->indexAtPlaylistIndex(viewState.topIndex, false);
+              }
+
+              if(allNew && !topIndex.isValid()) {
```
ludouzi: `allNew` means the entire playlist has changed e.g. a library viewer playlist. After this change, the view will incorrectly restore the "old" playlist scroll position, instead of starting at the top.

The `!topIndex.isValid()` condition should be removed here.

ludouzi (changes requested):

src/gui/playlist/editableplaylistsession.cpp
bassstorm: Thanks for review! Agreed and updated accordingly.
On a side note... Probably whole method needs refactoring to reduce complexity. I tried to keep to original code as close as possible :)

ludouzi (approved):
Thanks!

PR #1318, opened 2026-06-21
[core] Scripting: Add default arg to $elide_{mid,end}()

lstrsrt:
99% of the time, the `$elide` functions will be called with "..." as the delimiter, so default to it. Also update the reference entries, which did not mention the third argument (should we denote default/optional args differently?).


ludouzi:
> should we denote default/optional args differently?

We already denote optional args with brackets: `$if(condition,then[,else])`, so the elide entries should use the same format.

lstrsrt:
Ok, done. I also went through the other functions and added any missing brackets/args.

ludouzi (approved):
Thanks!

PR #1333, opened 2026-06-25
[gui] Improve properties dialog performance

lstrsrt:
This PR addresses some issues related to the properties dialog.

We currently rebuild the scope list every time `refreshScopePanel()` is called (which happens a lot, including on selection changes). This causes CPU spikes and freezes on my machine when scrolling through large (> 2000) track lists using arrow keys. Most of the time is spent doing metadata comparisons. So:

* Disallow duplicate track items.
* Add `buildScopePanel()` to only build the list on widget construction, and make
  `refreshScopePanel()` only refresh existing items. This also fixes shift + arrow key selections.
* Add a cache for pending track changes to the session and widget to avoid the metadata comparisons.

When the tag editor is enabled, it still takes some time to open the properties dialog with big track lists due to the info populator reading and adding track metadata. I don't think this can be avoided, but perhaps it could be done asynchronously?

I also found two bugs while testing that existed before this PR: editing the lyrics in the lyrics tab will mark the track as pending, but changing them back will not unmark them. Editing the RG data does not mark the track as pending at all.

### DIFF
--- a/include/gui/propertiesdialog.h
+++ b/include/gui/propertiesdialog.h
@@ -58,7 +58,7 @@ private:
     TrackList m_workingTracks;
     std::set<int> m_activeTrackIndexes;
     std::vector<int> m_pendingRevisions;
-    qsizetype m_pendingCount;
+    qsizetype m_pendingCount{0};
 };
 
 class FYGUI_EXPORT PropertiesTabWidget : public QWidget
--- a/include/gui/propertiesdialog.h
+++ b/include/gui/propertiesdialog.h
@@ -51,12 +51,14 @@ public:
 
     [[nodiscard]] bool hasChanges() const;
     [[nodiscard]] bool hasOnlyStatChanges() const;
-    [[nodiscard]] bool isTrackDirty(int index) const;
+    [[nodiscard]] int trackRevision(int index) const;
 
 private:
     TrackList m_originalTracks;
     TrackList m_workingTracks;
     std::set<int> m_activeTrackIndexes;
+    std::vector<int> m_pendingRevisions;
+    qsizetype m_pendingCount;
 };
 
 class FYGUI_EXPORT PropertiesTabWidget : public QWidget
--- a/src/gui/dialog/propertiesdialog.cpp
+++ b/src/gui/dialog/propertiesdialog.cpp
@@ -68,7 +68,7 @@ Fooyin::TrackList filterDuplicateTracks(const Fooyin::TrackList& tracks)
     filteredTracks.reserve(tracks.size());
 
     for(const auto& track : tracks) {
-        if(seenTracks.emplace(track.uniqueFilepath()).second) {
+        if(seenTracks.emplace(track.identityKey()).second) {
             filteredTracks.push_back(track);
         }
     }
@@ -100,7 +100,7 @@ void PropertiesDialogSession::reset(const TrackList& tracks)
     m_originalTracks         = workingTracks;
     m_workingTracks          = workingTracks;
     m_activeTrackIndexes.clear();
-    m_pendingRevisions.resize(static_cast<qsizetype>(workingTracks.size()));
+    m_pendingRevisions.assign(static_cast<qsizetype>(workingTracks.size()), 0);
     m_pendingCount = 0;
 }
 
--- a/src/gui/dialog/propertiesdialog.cpp
+++ b/src/gui/dialog/propertiesdialog.cpp
@@ -175,7 +175,7 @@ void PropertiesDialogSession::updateTracks(const TrackList& tracks)
             }
             m_pendingRevisions[index]++;
         }
-        else {
+        else if(m_pendingRevisions[index] > 0) {
             m_pendingRevisions[index] = 0;
             m_pendingCount            = std::max(qsizetype{0}, m_pendingCount - 1);
         }
--- a/src/gui/dialog/propertiesdialog.cpp
+++ b/src/gui/dialog/propertiesdialog.cpp
@@ -48,6 +48,7 @@
 
 #include <algorithm>
 #include <ranges>
+#include <unordered_set>
 #include <utility>
 
 using namespace Qt::StringLiterals;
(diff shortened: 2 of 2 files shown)


src/gui/dialog/propertiesdialog.cpp
```diff
@@ -712,44 +732,70 @@ void PropertiesDialogWidget::refreshScopePanel()
...
+    if(auto* allTracksItem = m_scopeList->item(0)) {
+        const bool markAllTracksPending = m_session.hasChanges() || pendingAllTracks || pendingAnyTracks;
+        allTracksItem->setText((markAllTracksPending ? u"* "_s : QString{}) + scopeLabel(-1));
+    }
+
+    const TrackList& tracks = m_session.workingTracks();
+
+    for(size_t i{0}; i < tracks.size(); ++i) {
+        const int trackIndex    = static_cast<int>(i);
+        const bool pendingTrack = hasPendingScopeChanges && m_session.activeTrackIndexes().contains(trackIndex);
```
lstrsrt, line 774: Ideally I would like to get rid of `pendingTrack` and just check the revision but doing it right now would mean that the star only appears on the next track update i.e. when scope is switched. I wonder if we can update the tracks every time a field is edited? That way `scopeLabel()` would also return up-to-date values.

src/gui/dialog/propertiesdialog.cpp
```diff
@@ -58,6 +59,23 @@ constexpr auto BaseWidthKey            = "PropertiesDialog/BaseWidth";
...
+Fooyin::TrackList filterDuplicateTracks(const Fooyin::TrackList& tracks)
+{
+    std::unordered_set<QString> seenTracks;
+    seenTracks.reserve(tracks.size());
+
+    Fooyin::TrackList filteredTracks;
+    filteredTracks.reserve(tracks.size());
+
+    for(const auto& track : tracks) {
+        if(seenTracks.emplace(track.uniqueFilepath()).second) {
```
ludouzi: `uniqueFilepath()` doesn't include `subsong()` so multiple subsongs will collapse to the first subsong. It would probably be best to use `identityKey()` after rebasing.

src/gui/dialog/propertiesdialog.cpp
ludouzi: A revision represents “something has changed” while the star is state (“currently dirty/pending”), so both concepts should ideally be kept.

Updating tracks on each edit would lead to every property tab being updated as well, potentially introducing another performance problem. It would also complicate reverting track edits to their original value. I would keep it as is for now.

We could look into adding a per-track presentation state to replace the revision/pending logic later, or make updating tracks incremental.

src/gui/dialog/propertiesdialog.cpp
```diff
@@ -78,9 +96,12 @@ bool editorEqual(const Fooyin::Track& lhs, const Fooyin::Track& rhs)
 namespace Fooyin {
 void PropertiesDialogSession::reset(const TrackList& tracks)
 {
-    m_originalTracks = tracks;
-    m_workingTracks  = tracks;
+    const auto workingTracks = filterDuplicateTracks(tracks);
+    m_originalTracks         = workingTracks;
+    m_workingTracks          = workingTracks;
     m_activeTrackIndexes.clear();
+    m_pendingRevisions.resize(static_cast<qsizetype>(workingTracks.size()));
```
ludouzi: `resize()` preserves elements when the session is reset to the same or a smaller size, so `trackRevision()` will contain stale info.

This should use assign: `m_pendingRevisions.assign(workingTracks.size(), 0);`

include/gui/propertiesdialog.h
```diff
@@ -51,12 +51,14 @@ class FYGUI_EXPORT PropertiesDialogSession
...
     [[nodiscard]] bool hasOnlyStatChanges() const;
-    [[nodiscard]] bool isTrackDirty(int index) const;
+    [[nodiscard]] int trackRevision(int index) const;
 
 private:
     TrackList m_originalTracks;
     TrackList m_workingTracks;
     std::set<int> m_activeTrackIndexes;
+    std::vector<int> m_pendingRevisions;
+    qsizetype m_pendingCount;
```
ludouzi: This should be default-initialised to 0.

ludouzi:
> When the tag editor is enabled, it still takes some time to open the properties dialog with big track lists due to the info populator reading and adding track metadata. I don't think this can be avoided, but perhaps it could be done asynchronously?

Yeah, it could easily be done similar to the selection info widget which is loaded asynchronously. The tag editor can also be embedded as a layout widget, which means selecting all tracks in a library can block the main thread for a time as well. I'll look into it.

> I also found two bugs while testing that existed before this PR

Thanks! I'll look into those too.

src/gui/dialog/propertiesdialog.cpp
lstrsrt: Actually the way it works now, the star is added immediately on editing a field, so that part of my comment doesn't apply. We do make `pendingTrack` increment the revision which violates the separation a little. By incremental, do you mean updating only tracks that were edited? This is what I also had in mind.

lstrsrt:
> Yeah, it could easily be done similar to the selection info widget which is loaded asynchronously. The tag editor can also be embedded as a layout widget, which means selecting all tracks in a library can block the main thread for a time as well. I'll look into it.

Great 👍

Rebased and addressed review.

src/gui/dialog/propertiesdialog.cpp
ludouzi: Yeah, that's what I had in mind.

lstrsrt: I think this would involve changes to the tag editor plugin. I would open a new PR for this so as to not overcomplicate this one.

src/gui/dialog/propertiesdialog.cpp
```diff
@@ -147,28 +168,31 @@ void PropertiesDialogSession::updateTracks(const TrackList& tracks)
...
+        const auto index = std::distance(m_workingTracks.begin(), workingIt);
+        if(!editorEqual(m_originalTracks[index], updatedTrack)) {
+            if(m_pendingRevisions[index] == 0) {
+                m_pendingCount++;
+            }
+            m_pendingRevisions[index]++;
+        }
+        else {
+            m_pendingRevisions[index] = 0;
+            m_pendingCount            = std::max(qsizetype{0}, m_pendingCount - 1);
```
ludouzi, line 180: Clean/non-edited tracks can decrement the pending count here as updates include unchanged tracks, so hasChanges() can incorrectly become false.

lstrsrt: I believe this is now fixed, but I wasn't able to test it. Is there a way to trigger such an update? In my testing `updateTracks()` is only called after switching scope with a pending edit, and only receives the tracks that were updated.

ludouzi: The path I had in mind was: `commitCurrentScopeEdits()` calls `TagEditorEditor::applyChanges()`, which returns `m_model->tracks()`, which is the entire active scope.

That could be triggered by:

1. Open properties with multiple tracks
2. Keep the default `All Tracks` scope
3. Edit a field that only changes some tracks
4. Switch scope

That should now be fixed.

ludouzi (approved):

PR #1340, opened 2026-06-28
Confirm before removing a playlist with a don't-ask-again option

### LINKED ISSUE #1301: Make it harder to delete playlists
At present, a playlist can be deleted simply by hitting the Delete key, and there is no undo. It can be that you think you're going to remove a track from a playlist, but if the focus happens to be in the playlist organizer, suddenly the whole playlist is permanently gone.

It would be wonderful if it were just a bit harder to delete. Such as:

- No playlist deletion simply by hitting Delete.
- Right-click and choose **Delete Playlist**.
- Then confirmation: "**Really delete playlist? (Yes) (Cancel)**"

mvanhorn:
## Summary

Deleting a playlist now shows a Yes/No confirmation dialog before it is removed, with a "Don't ask again" checkbox so users who don't want the prompt can opt out. This guards against accidental one-click deletions, which is the case raised in the issue (a shared setup where operators removed playlists without realising it).

## Why this matters

Every playlist removal path called `PlaylistHandler::removePlaylist` directly with no guard, so a single context-menu click, a middle-click on a tab, the close button, or the Ctrl+W shortcut would permanently drop a playlist. The confirmation is routed through one shared helper (`confirmPlaylistRemoval`) so all of these entry points behave consistently:

- `PlaylistBox::removeCurrentPlaylist` (src/gui/playlist/playlistbox.cpp)
- `PlaylistManagerWidget::removePlaylist` (src/gui/playlist/manager/playlistmanagerwidget.cpp)
- `PlaylistTabs` context-menu remove, middle-click close, and tab-close (src/gui/playlist/playlisttabs.cpp)
- The global "Remove Playlist" action / Ctrl+W (src/gui/guiapplication.cpp)
- The playlist organiser remove action, which only prompts when the selection actually contains playlists (src/gui/playlist/organiser/playlistorganiser.cpp)

The preference is stored as a new `GuiSettings::PlaylistRemovalConfirm` boolean (key `Playlist/RemovalConfirm`, default on), registered in src/gui/internalguisettings.cpp. Ticking "Don't ask again" and confirming sets it to false, after which removals skip the dialog. Autoplaylists share the same paths, so they are covered too.

## Testing

Qt6 is not available in my build environment, so I verified by hand and with clang-format rather than a full build:

- `clang-format --dry-run --Werror` passes on all changed files against the repo `.clang-format`.
- Confirmed the new setting reads/writes use the existing templated `settings->value<...>()` / `settings->set<...>()` API and the registration matches the surrounding `createSetting<Name>(default, u"Group/Key"_s)` style.
- Grepped every `removePlaylist` entry point to confirm each now routes through `confirmPlaylistRemoval`.

Fixes #1301

### DIFF
--- a/include/gui/guisettings.h
+++ b/include/gui/guisettings.h
@@ -91,6 +91,7 @@ enum GuiSettings : uint32_t
     PlaylistSearchMode         = 34 | Type::Int,
     PlaylistSearchScript       = 35 | Type::String,
     DragOnlyAfterSelect        = 36 | Type::Bool,
+    PlaylistRemovalConfirm     = 37 | Type::Bool,
 };
 Q_ENUM_NS(GuiSettings)
 } // namespace Settings::Gui
--- a/src/gui/guiapplication.cpp
+++ b/src/gui/guiapplication.cpp
@@ -36,6 +36,7 @@
 #include "menubar/playbackmenu.h"
 #include "menubar/viewmenu.h"
 #include "playlist/manager/playlistmanagerwidget.h"
+#include "playlist/playlistbox.h"
 #include "playlist/playlistcontroller.h"
 #include "playlist/playlistinteractor.h"
 #include "playlist/playlistuicontroller.h"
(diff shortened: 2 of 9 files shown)


ludouzi:
Closing this as the issue has been handled in 0ba3a666345a796370ae5f4845124c77e802e387.

PR #1343, opened 2026-06-29
fix: preserve literal < and > in playlist script output

### LINKED ISSUE #1224: some special characters can't be displayed
**Describe the bug**
some special characters can't be displayed it the playlist. even if I put quotation marks around the characters.
the characters are:  <  and  >
by the way in foobar2000 I used ' ' as quotation marks but fooyin the correct way seems to be to use " "

**Expected behaviour**
the characters should be displayed when quotation marks are present

**System:**
 - OS: openSUSE tumbleweed
 - fooyin version: 0.10.8

mvanhorn:
## Summary
Literal `<` and `>` characters now display in the playlist instead of being silently dropped. A title like `<3` or `a > b` renders as written, with no need for the `\<` escape workaround.

## Why this matters
The playlist runs script output through a second formatting pass, and `ScriptFormatter` treated every `<...>` sequence as a formatting tag (issue #1224). Unrecognised angle-bracket content was consumed and discarded. In `src/gui/scripting/scriptformatter.cpp`, `formatBlock()` now only enters tag processing when the text after `<` is a recognised formatter (via `ScriptFormatterRegistry::isKnown`) that closes with `>`; otherwise it emits the original `<...>` text verbatim. `readTagContent()` stops at a nested `<` so a literal `<` cannot swallow a following real tag, and angle detection checks the token value so the scanner keywords `LESS`/`GREATER` are not mistaken for bracket characters. Known tags with bad options still fall through to `processFormat()` and render their inner text.

## Testing
Added cases to `tests/gui/scriptformattertest.cpp` covering literal brackets with spaces (`a < b > c`), `I <3 this`, an unknown tag kept literal (`<notatag>rest`), literal text mixed with real formatting (`I <3 <b>this</b>` and `<b>I <3 this</b>`), and the keyword words (`<LESS>`, `5 LESS than 6`). Existing bold/italic/colour/link tests continue to assert no regression. The Qt6 toolchain is not installed in this environment, so the suite was not run here; clang-format passes on both files.

Fixes #1224

### DIFF
--- a/src/gui/scripting/scriptformatter.cpp
+++ b/src/gui/scripting/scriptformatter.cpp
@@ -180,10 +180,30 @@ void ScriptFormatterPrivate::errorAt(const ScriptScanner::Token& token, const QS
     m_errors.emplace_back(currentError);
 }
 
+namespace {
+// The formatter shares the query scanner, which also emits TokLeftAngle/TokRightAngle
+// for the keyword words "LESS"/"GREATER". A real bracket delimiter has the bracket
+// character as its value, so check the value rather than the type alone.
+bool isLiteralAngle(const ScriptScanner::Token& token, QChar bracket)
+{
+    return token.value.size() == 1 && token.value.front() == bracket;
+}
+
+bool isLeftAngle(const ScriptScanner::Token& token)
+{
+    return token.type == ScriptScanner::TokLeftAngle && isLiteralAngle(token, u'<');
+}
+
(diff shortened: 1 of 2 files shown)


ludouzi:
Closing this as 99afa811916f6d0a1db6a186809c28f5341a7529 replaces `ScriptScanner` with a simplified, dedicated cursor, which resolves #1224 and any similar issues going forward.

PR #1356, opened 2026-07-03
[tageditor] Fix renaming extended tags

lstrsrt:
This patch fixes tag renames in the tag editor not being saved.

We were using `node.field()` instead of the modified `field` for the `updateTracksMetadata()` call, which led to the original tag name being saved instead. We also weren't showing the value of the renamed field until the tab was switched or the widget was reopened.


ludouzi (approved):
Thanks. I fixed a small issue where renaming a tag blanked the displayed value until reopen. Looks good now!

PR #1367, opened 2026-07-08
Fix weird keyboard behaviour in fileops delete dialogue

numerfolt:
When deleting a file, a dialogue opens. In this dialogue the "Cancel" button should be focused, as it is highlighted.
The "Confirm"-Button is left to the "Cancel"-button. 

But when pressing the left arrow to focus "Confirm", the focus first switches to "Cancel" although it should already be on it. Only after pressing the left arrow again, it focuses the "Confirm" button.
This behaviour is quite annoying and unintuitive.

Please see these pictures that were taken after opening the dialogue and pressing the left arrow button exactly once:
|Before the fix|After the fix|
|---|---|
|grafik|grafik|

(Please disregard the language)
What's important is, that while pressing the left arrow once in the current release only puts the focus on "Cancel", it puts the focus on "Confirm" with the fix.


ludouzi (approved):
Looks good, thanks!

PR #1369, opened 2026-07-09
[core] Fix Stop After Current when in deferred metadata flush mode

bassstorm:
When Stop After Current is enabled and there's pending metadata changes for current track, the track cursor doesn't advance forward upon playback stop. This PR schedules next track for proper pick-up.

### DIFF
--- a/src/core/playback/playbackordernavigator.cpp
+++ b/src/core/playback/playbackordernavigator.cpp
@@ -270,10 +270,13 @@ std::optional<PlaybackOrderNavigator::RequestedTrack> PlaybackOrderNavigator::se
     return {};
 }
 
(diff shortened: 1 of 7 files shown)


ludouzi (approved):

ludouzi:
Thanks!

Scheduling the resume track fixed the deferred metadata case, but gave it normal schedule priority, which caused `Next` and newly queued tracks to select the resume track first.

I fixed that by tagging the scheduled track with its purpose.

bassstorm:
I knew I missed something! I'd probably better just file a bug report next time :sweat_smile:
Nevertheless, glad the fix will be in next release. I use Stop After Current all the time, so having it broken was a bit annoying.

PR #1387, opened 2026-07-19
[tageditor] Fix new extended tags being duplicated by a rename

lstrsrt:
This patch just fixes one edge case I'd missed in my last PR, which is that when you add an extended tag, apply, rename it and apply again without closing the tag editor inbetween, two tags get saved: the original and the renamed one. The cause is that we set the script field to be updated to the key from `m_tags`, which is still the default field text at this point, so we clear a non-existent tag and never get rid of the original one.


ludouzi (approved):
Thanks!

PR #1388, opened 2026-07-19
Include instructions to test translation in README.md

numerfolt:
Up until now it isn't really clear to new contributors (like I was a year ago) how to test translations.

I created an issue last year and just now updated it to include instructions on how to test their translation.
Including all instructions in the README would be a bit much I think…


ludouzi (approved):

ludouzi:
Thanks!

I'll probably write a separate document for translations at some point, but this is a good start for now.

PR #1406, opened 2026-07-26
[sleepinhibitor] Support display idle inhibition

lstrsrt:
A range of visualizers is available now, so users might want to inhibit display sleep. Supporting this is relatively easy in most cases, just a matter of adding some flags to API calls. Notes:

* The D-Bus `org.freedesktop.PowerManagement.Inhibit` interface doesn't support display idle inhibition, so we have to add and track a second inhibitor on `org.freedesktop.ScreenSaver`. It's kind of hacky with the current design but it does work. We treat errors on that interface separately and still try to inhibit system suspend if we encounter one.
* There are races with disabling + re-enabling inhibition asynchronously (such as on D-Bus) so I added `m_desiredState` and `m_desiredType` to deal with them.
* The macOS thread commit is motivated by [this](https://source.chromium.org/chromium/chromium/src/+/main:services/device/wake_lock/power_save_blocker/power_save_blocker_mac.cc;l=22-25). It does look like newer versions of macOS try to [create assertions asynchronously](https://github.com/apple-oss-distributions/IOKitUser/blob/IOKitUser-100231.120.3/pwr_mgt.subproj/IOPMAssertions.c#L814) but there are many conditions that can let that function fail so let's be safe.

I know 0.12 is landing very soon so if it's too late for enhancement work I'm fine with delaying this patch.

Tested on GNOME, KDE and macOS Tahoe (VM).

### DIFF
--- a/src/plugins/sleepinhibitor/platforms/inhibitordbus.cpp
+++ b/src/plugins/sleepinhibitor/platforms/inhibitordbus.cpp
@@ -87,16 +87,23 @@ InhibitorDbus::InhibitorDbus(QObject* parent)
 
 void InhibitorDbus::inhibitScreenSaver()
 {
-    if(m_screenSaverError) {
+    m_screenSaverState.desiredState = State::Inhibited;
+
+    if(m_screenSaverState.currentState != State::Initializing
+       && m_screenSaverState.currentState != State::Uninhibited) {
         return;
     }
+
+    qCDebug(SLEEPINHIBITOR) << "Inhibiting ScreenSaver";
+    m_screenSaverState.currentState = State::Uninhibiting;
 
     if(!m_screenSaverInterface) {
         m_screenSaverInterface = new QDBusInterface(
             DbusConstants::FreedesktopScreenSaverService, DbusConstants::FreedesktopScreenSaverPath,
             DbusConstants::FreedesktopScreenSaverInterface, QDBusConnection::sessionBus(), this);
         if(!m_screenSaverInterface->isValid()) [[unlikely]] {
-            m_screenSaverError = true;
+            qCWarning(SLEEPINHIBITOR) << "Could not get usable ScreenSaver D-Bus interface";
+            m_screenSaverState.currentState = State::Error;
             return;
         }
     }
@@ -109,9 +116,14 @@ void InhibitorDbus::inhibitScreenSaver()
 
 void InhibitorDbus::uninhibitScreenSaver()
 {
-    if(m_screenSaverError || !m_screenSaverInhibitCookie) {
+    m_screenSaverState.desiredState = State::Uninhibited;
+
+    if(m_screenSaverState.currentState != State::Uninhibited || !m_screenSaverInhibitCookie) {
         return;
     }
+
+    qCDebug(SLEEPINHIBITOR) << "Uninhibiting ScreenSaver";
+    m_screenSaverState.currentState = State::Uninhibiting;
 
     const auto pendingCall = m_screenSaverInterface->asyncCall("UnInhibit"_L1, m_screenSaverInhibitCookie);
     auto* watcher          = new QDBusPendingCallWatcher(pendingCall, this);
@@ -125,11 +137,17 @@ void InhibitorDbus::onScreenSaverInhibitCallFinished(QDBusPendingCallWatcher* wa
     watcher->deleteLater();
 
     if(reply.isValid()) {
-        m_screenSaverInhibitCookie = reply.value();
+        m_screenSaverInhibitCookie      = reply.value();
+        m_screenSaverState.currentState = State::Inhibited;
     }
     else {
         qCWarning(SLEEPINHIBITOR) << "ScreenSaver Inhibit call error:" << reply.error().message();
-        m_screenSaverError = true;
+        m_screenSaverState.currentState = State::Error;
+    }
+
+    if(m_screenSaverState.desiredState == State::Uninhibited) {
+        uninhibitScreenSaver();
+        m_screenSaverState.desiredState = {};
     }
 }
 
@@ -139,24 +157,31 @@ void InhibitorDbus::onScreenSaverUninhibitCallFinished(QDBusPendingCallWatcher* 
     watcher->deleteLater();
 
     if(reply.isValid()) {
-        m_screenSaverInhibitCookie = 0;
+        m_screenSaverInhibitCookie      = 0;
+        m_screenSaverState.currentState = State::Uninhibited;
     }
     else {
         qCWarning(SLEEPINHIBITOR) << "ScreenSaver Uninhibit call error:" << reply.error().message();
-        m_screenSaverError = true;
+        m_screenSaverState.currentState = State::Error;
+    }
+
+    if(m_screenSaverState.desiredState == State::Inhibited) {
+        inhibitScreenSaver();
+        m_screenSaverState.desiredState = {};
     }
 }
 
 void InhibitorDbus::inhibitSleep(InhibitionType type)
 {
-    m_desiredState = State::Inhibited;
-    m_desiredType  = type;
+    m_powerState.desiredState = State::Inhibited;
+    m_powerState.desiredType  = type;
 
-    if(state() == State::Error || state() == State::Inhibited) {
+    if(state() != State::Initializing && state() != State::Uninhibited) {
         return;
     }
 
     qCDebug(SLEEPINHIBITOR) << "Inhibiting sleep";
+    setState(State::Inhibiting);
 
     static constexpr auto BlockLogoutFlag  = 1U;
     static constexpr auto BlockSuspendFlag = 4U;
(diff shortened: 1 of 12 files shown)


bassstorm:
Yes! I don't use visualizers, but noticed my KDE system 10m idle timeout puts display to sleep, while I was actually expecting **sleep** inhibitor plugin to prevent that while Fooyin is playing :) Appreciate you contribution and hope to see this improvement sooner than later.

Update:
An important detail, above happens despite KDE showing presentation enforcement by Fooyin.
image
The display doesn't go to sleep when I enable KDE's presentation mode toggle.

lstrsrt:
KDE seems to consider presentation mode "enforced" [when any kind of inhibitor is active](https://github.com/KDE/kscreen/blob/master/plasmoid/PresentationModeItem.qml#L68). The battery applet is more specific here, and with this patch you should see the following:

Screenshot From 2026-07-28 14-34-16

bassstorm:
Thanks for hint, I apparently never put enough attention to Power and Battery widget. However, it's already shown like that on my system with Fooyin 0.12.0...

image

Hopefully this patch will add missing glue to make the feature work universally.

P.S. I'll test the patch when it hits mainstream. Sorry for being lazy to set up one more fork instance :)

lstrsrt:
Ah I see. I think you are on an older KDE version than the one I tested with (6.7.3). [This commit](https://github.com/KDE/powerdevil/commit/83f685b79cc31fb5aae03c16285da3f05907a47d#diff-995981984d0231e20002da767bbd229a13eafaee6edea60e9cd2555bdb36fb50L145-R183) to powerdevil from 6 months ago changed the way the applet handles inhibitors. On your end it shows up as "sleep and screen locking" because the older code did not distinguish between inhibition types.

Also given that you can unblock inhibition directly from the applet UI, I now wonder if we should relax error handling a bit and allow our uninhibit functions to fail. Then inhibition could be unblocked externally without us giving up entirely.

bassstorm:
I certainly am :) Precisely, Debian Stable and KDE 6.3.6. I can test some edge cases on my system as well when time comes if you think that would be helpful. Just lemme know.

src/plugins/sleepinhibitor/platforms/inhibitormacos.cpp
```diff
@@ -18,49 +18,73 @@
...
-    }
+        const auto assertionName = tr("fooyin is running").toCFString();
+        const auto assertionType = type == InhibitionType::DisplayAndSystem
+                                     ? kIOPMAssertionTypePreventUserIdleDisplaySleep
+                                     : kIOPMAssertionTypePreventUserIdleSystemSleep;
+        const auto status
+            = IOPMAssertionCreateWithName(assertionType, kIOPMAssertionLevelOn, assertionName, &m_assertionId);
+        CFRelease(assertionName);
+        return status;
+    }).then([this](IOReturn status) {
```
ludouzi: This `QFuture::then()` call should pass `this` as a context object, otherwise it's not thread safe as it doesn't run in the main thread.

lstrsrt: Done

lstrsrt:
Build failures unrelated I think.

src/plugins/sleepinhibitor/platforms/inhibitormacos.cpp
```diff
@@ -18,49 +18,73 @@
...
 {
+    m_desiredState = State::Inhibited;
+    m_desiredType  = type;
+
     if(state() == State::Error || state() == State::Inhibited) {
         return;
     }
 
-    qCDebug(SLEEPINHIBITOR) << "Inhibiting sleep";
+    QtConcurrent::run([this, type] -> IOReturn {
```
ludouzi, line 39: `QtConcurrent::run()` shouldn’t access or mutate `m_assertionId` directly. The worker can outlive the inhibitor object, and during plugin shutdown the `uninhibitSleep()` worker has the same problem.

Chromium avoids this by keeping the assertion in a delegate bound to one dedicated sequence. Maybe we could similarly use a dedicated worker that owns the assertion, then drain/join it before unloading the plugin. This would also prevent concurrent create/release operations from racing over the assertion ID.

src/plugins/sleepinhibitor/platforms/inhibitordbus.cpp
```diff
@@ -76,13 +79,79 @@ InhibitorDbus::InhibitorDbus(QObject* parent)
...
+}
+
+void InhibitorDbus::inhibitScreenSaver()
+{
+    if(m_screenSaverError) {
+        return;
+    }
+
+    if(!m_screenSaverInterface) {
+        m_screenSaverInterface = new QDBusInterface(
```
ludouzi, line 116: This needs a `QObject` parent.

src/plugins/sleepinhibitor/platforms/inhibitordbus.cpp
```diff
@@ -76,13 +79,79 @@ InhibitorDbus::InhibitorDbus(QObject* parent)
...
+    }
 }
 
-void InhibitorDbus::inhibitSleep()
+void InhibitorDbus::inhibitSleep(InhibitionType type)
 {
+    m_desiredState = State::Inhibited;
+    m_desiredType  = type;
+
     if(state() == State::Error || state() == State::Inhibited) {
```
ludouzi: The state is not changed while this async inhibit call is pending, so another `inhibitSleep()` can launch a second request before the first completes e.g. play -> pause -> play during the initial request causes `uninhibitSleep()` to return early and then starts another inhibit request. Both replies overwrite the same cookie/handle, leaving one inhibitor that can never be released.

We should probably represent `Inhibiting/Uninhibiting` explicitly, allow only one operation at a time, and resolve the latest desired state/type after each completion.

src/plugins/sleepinhibitor/platforms/inhibitordbus.cpp
```diff
@@ -76,13 +79,79 @@ InhibitorDbus::InhibitorDbus(QObject* parent)
...
+                     &InhibitorDbus::onScreenSaverUninhibitCallFinished);
+}
+
+void InhibitorDbus::onScreenSaverInhibitCallFinished(QDBusPendingCallWatcher* watcher)
+{
+    const QDBusPendingReply<uint> reply = *watcher;
+    watcher->deleteLater();
+
+    if(reply.isValid()) {
+        m_screenSaverInhibitCookie = reply.value();
```
ludouzi: This late reply is not reconciled with the desired state. If the PowerManagement request completes first and playback then stops, `uninhibitScreenSaver()` sees a zero cookie and returns; this callback then stores the cookie, leaving inhibition active. An overlapping `UnInhibit` reply can also clear the cookie from a newer inhibit request.

The ScreenSaver inhibitor needs its own serialised pending/actual/desired state so a late successful inhibit is immediately released when no longer wanted.

lstrsrt:
Thanks for the review @ludouzi, I started a new full time position this month so it has been hard to do any open source work. I'll try to get to it on the weekend.

ludouzi:
No worries! Best of luck in the new role. I’ll be glad to help out on this if needed.

src/plugins/sleepinhibitor/platforms/inhibitordbus.cpp
lstrsrt: Does the same apply to `inhibitRequestInterface`?

src/plugins/sleepinhibitor/platforms/inhibitordbus.cpp
lstrsrt: First two points should be handled with `6460291`. "resolve the latest desired state/type after each completion" -- isn't this what we're doing now (checking desired state in the async callback), or something else?

src/plugins/sleepinhibitor/platforms/inhibitormacos.cpp
lstrsrt: With the current design, the internals of `InhibitorMacOs` need to be accessed and changed based on what syscall return values we get, what our desired state is and so on. I am not sure how to do this if all the work is moved to a new worker class.

bassstorm:
Hi folks,
I see some post-review updates followed by silence. Any chance to see this improvement in mainline?

ludouzi:
It needs another review and some changes. I'll get around to it within the next few days.

src/plugins/sleepinhibitor/platforms/inhibitordbus.cpp
ludouzi: Not needed for `inhibitRequestInterface`. Though the pending call is still valid after the interface goes out of scope, so I moved it to the stack.

ludouzi (approved):

ludouzi:
Thanks - this looks ready to merge now. 

There are probably still some edge cases left, particularly around users unblocking inhibition outside fooyin, but we can test and handle those separately :)

PR #1557, opened 2026-08-26
[core] TagLib: Flush file device after saving

lstrsrt:
Problem: atime and mtime timestamps weren't being preserved on tag updates when "Preserve timestamps" was enabled on my system. This is caused by the file device being flushed too late which discards the timestamp restoration.

Fix: Call `flush()` on the file device after saving the file to force changes to be written to disk before restoring timestamps. Note that atime timestamp preservation is still broken for me, probably due to surrounding `QFileInfo` and `IODeviceStream` constructors which have to access the file. In my opinion preserving atime is overkill but I'll leave it up to you whether it should be fixed or removed.

### DIFF
--- a/src/core/engine/input/taglibparser.cpp
+++ b/src/core/engine/input/taglibparser.cpp
@@ -3911,14 +3911,17 @@ bool TagLibReader::writeTrack(const AudioSource& source, const Track& track, Wri
         return false;
     }
 
-    if(atime.isValid() || mtime.isValid()) {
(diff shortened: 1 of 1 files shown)


ludouzi (approved):

ludouzi:
Thanks!

> Note that atime timestamp preservation is still broken for me

Perhaps that is due to setting the mtime after the atime? I'll play around with it and see if I can reproduce it with tests.

lstrsrt:
The test added in `df2c061` passes on my end, but for actual music files it is still not working.

PR #1558, opened 2026-08-27
[core,gui,cdda] Add Audio CD playback and ripping

### LINKED ISSUE #834: Add CD Audio support?
Cant seem to play CD discs as I think it uses the Dolphin KIO AudioCD which only seems to rip the cds, not play the disc directly.

I tried disabling the KIO to see if it had a built in reader but Fooyin doesnt recognise the CD.

ludouzi:
Adds functionality to open and read audio CDs, including playback and ripping using the existing metadata lookup and converter workflows.

A new `File -> Open audio CD` menu option opens the initial dialog. This supports reading CD-Text and looking up metadata before adding the CD tracks to a playlist and/or playing.
image

Each drive's individual settings used for ripping can also be configured:
image

Using the `Rip` button opens the `Rip Audio CD` dialog, from which metadata can be edited and looked up again. The ripping process uses the saved converter presets, or the converter can be opened directly.

image

If `Verify with AccurateRip` is checked, a verification report will be shown at the end of the ripping process:

image

Closes #834

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -30,6 +30,7 @@ The following libraries are optional:
 * [Game Music Emu](https://github.com/libgme/game-music-emu) - for the GME audio input plugin
 * [libarchive](https://www.libarchive.org) - for the archive support plugin
 * [libebur128](https://github.com/jiixyj/libebur128) - for the ReplayGain scanner plugin
+* [libcdio](https://www.gnu.org/software/libcdio/) and libcdio-paranoia - for the Audio CD plugin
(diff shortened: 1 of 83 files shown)


dustcap:
Thank you for the CD ripping support!

Actually it doesn't work with Windows, or is Win10 the problem?

ludouzi:
This will be Linux-only for now, as the library we're using doesn't fully support MSVC on Windows yet:

https://github.com/libcdio/libcdio-paranoia/issues/63

Plus it's not in any Windows-based package managers, so we'll need to build and package it ourselves somewhere.

PR #1580, opened 2026-09-03
[gui] Typography touch-ups no. 2

ElectronicFate:
- Replaced some more hyphens that were used as a pause with em dashes.
- Removed spacing around en dashes used for ranges of values.
- Replaced ellipses that used three individual dots with ellipses.


ludouzi (approved):
Thanks!

PR #1582, opened 2026-09-05
build: delete extraneous dependencies

jengelh:
fooyin's cmake files erroneously request packages whose header files it never uses. Turn that off.
(In addition, postproc is also gone in ffmpeg-9.)


ludouzi (approved):
Thanks!

PR #1625, opened 2026-09-16
[tageditor] Fix crash when heart delegate is null

lstrsrt:
If the rating editor is enabled but the heart editor is not, the tag editor will crash in `mouseMoveEvent` since `m_starDelegate` exists but `m_heartDelegate` does not, so the latter gets accessed without a null check. Fix: Add the check.


ludouzi (approved):
Oops! Thanks!

